PR #4, opened 2023-06-28
Time domain 

charlotteaward:
Adding channel mapping to renderer plus static source and static spectrum capabilities

### DIFF
--- a/scarlet2/observation.py
+++ b/scarlet2/observation.py
@@ -91,10 +91,10 @@ class Observation():
         return self
     
     # Matt addition, make __eq__ and __hash__ method for JIT compile to make obs hashable
-    # def __eq__(self, other):
-    #     # Equality Comparison between two objects
+    #def __eq__(self, other):
+         # Equality Comparison between two objects
     #     return self.data == other.data and self.weights == other.weights and self.frame == other.frame and self.renderer == other.renderer 
 
-    # def __hash__(self):
-    #     # hash(custom_object)
+    #def __hash__(self):
+         # hash(custom_object)
     #     return hash((self.data))
--- a/scarlet2/renderer.py
+++ b/scarlet2/renderer.py
@@ -4,10 +4,6 @@ from .fft import convolve, deconvolve, _get_fast_shape
 from .module import Module
 
 class Renderer(Module):
-    def __init__(self, frame, obs_frame, *parameters):
-        object.__setattr__(self, "channel_map", self.get_channel_map(frame, obs_frame))
-        #self.channel_map = self.get_channel_map(frame, obs_frame)
-        super().__init__(*parameters)
     def __call__(self, model):
         raise NotImplementedError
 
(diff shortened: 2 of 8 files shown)


scarlet2/observation.py
pmelchior, line 1: These changes should not have been committed, they're just whitespace edits. What editor did you use that modified these?

scarlet2/renderer.py
```diff
@@ -3,28 +3,61 @@
 from .fft import convolve, deconvolve, _get_fast_shape
 from .module import Module
 
-
 class Renderer(Module):
+    def __init__(self, frame, obs_frame, *parameters):
+        object.__setattr__(self, "channel_map", self.get_channel_map(frame, obs_frame))
+        #self.channel_map = self.get_channel_map(frame, obs_frame)
```
pmelchior, line 9: please remove

scarlet2/renderer.py
```diff
@@ -3,28 +3,61 @@
...
+            The hyperspectral model
+        Returns
+        -------
+        obs_model: array
+            `model` mapped onto the observation channels
+        """
+        if self.channel_map is None:
+            return model
+        if isinstance(self.channel_map, slice):
+            return model[self.channel_map]
```
pmelchior, line 41: we may have to fix this for jit, but we can fix it when it becomes a problem

scarlet2/source.py
```diff
@@ -31,6 +31,35 @@ def __call__(self):
     def bbox(self):
         return self.spectrum.bbox @ self.morphology.bbox
 
+class StaticSource(Module):
+    spectrum: Spectrum
+    morphology: Morphology 
+    specind: Specind
```
pmelchior, line 37: What's the purpose of `Specind`? It looks to be a simple box, which you can set for the spectrum already.

pmelchior: Also, it's better to use class names that are CamelCase and descriptive, so that we don't have to guess what the class does.

scarlet2/spectrum.py
```diff
@@ -1,12 +1,21 @@
...
     def __init__(self, *args, **kwargs):
         Parameter.__init__(self, *args, **kwargs)
-        self.bbox = Box(self.value.shape)
\ No newline at end of file
+        self.bbox = Box(self.value.shape)
+
+class StaticArraySpectrum(Spectrum, Parameter):
+    def __init__(self, *args, **kwargs):
+        Parameter.__init__(self, *args, **kwargs)  
+        self.bbox = Box(jnp.repeat(self.value,self.repeats).shape)
```
pmelchior, line 20: I'm confused: 
* I don't see a definition of `self.repeats`.
* How does manipulating the `bbox` get you a source that has the same spectrum across multiple observations/channels. I think this is related to the `specind` situation above, where you seem to multiply the spectra with a slicing operation.

pmelchior:
closed in favor of #6

PR #6, opened 2023-07-10
Adding static source and channel mapping

charlotteaward:


scarlet2/spectrum.py
```diff
@@ -27,11 +27,11 @@ class StaticArraySpectrum(Spectrum):
     def __init__(self, data, filters, scene):
```
pmelchior, line 27: One other minor issue: please add ` = eqx.field(static=True)` as the default for `channelindex`.

pmelchior (approved):

PR #13, opened 2023-09-28
neural prior class addition (requires galaxygrad==0.0.19)

SampsonML:


scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+# This class is used to calculate the gradient       #
+# of the log-probability via calling the nn prior    #
+# ScoreNet model. A custom vjp is created to return  #
+# the nn prior when calling jax.grad()               #
+# -------------------------------------------------- #
+
+try:
+    import numpyro.distributions as dist
+    import numpyro.distributions.constraints as constraints
+    from galaxygrad import HSC_ScoreNet32, HSC_ScoreNet64, ZTF_ScoreNet32, ZTF_ScoreNet64,HSC_LogNet32, HSC_LogNet64# (https://pypi.org/project/galaxygrad/0.0.19)
```
pmelchior, line 11: I don't think this should be a required import. This class needs to define the linkage between galaxygrad and scarlet2, but a user should import whatever model from galaxygrad they want to use.

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+except ImportError:
+    raise ImportError("scarlet2.nn requires numpyro and galaxygrad=0.0.19")
+
+import jax.numpy as jnp
+import jax.scipy as jsp
+from jax import custom_vjp
+from jax import vjp
+
+
+def pad_fwd(x, trained_model):
```
pmelchior, line 22: `trained_model` should not omitted here. It's used to get the model from its name given the data shape, but when the user sets the specific galaxygrad model for a specific source size already, this isn't necessary.

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+    elif trained_model == 'ztf':
+        if data_size <= 32:
+            pad_gap = 32 - data_size
+            ScoreNet = ZTF_ScoreNet32
+        else:
+            pad_gap = 64 - data_size
+            ScoreNet = ZTF_ScoreNet64
+    # select the custom trained model
+    else:
+        ScoreNet = trained_model
```
pmelchior, line 47: Without the model selection, this is much easier: if the shape is smaller than the max_shape, pad it.

scarlet2/nn.py
pmelchior: Instead, pass a `max_size` argument, which determines the full image size that was used for the score model.

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+def pad_back(x, pad_lo, pad_hi):
+    """Removes the zero-padding from the input image"""
+    if jnp.ndim(x) > 2: 
+        x[0] = x[pad_lo:-pad_hi, pad_lo:-pad_hi]
+    else:
+        x = x[pad_lo:-pad_hi, pad_lo:-pad_hi]
+    return x
+
+# calculate score function (jacobian of log-probability)
+def calc_grad(x, trained_model):
```
pmelchior, line 75: let's rename `trained_model` with `model`, by which we mean the actual jax model that computes the score (instead of its name)

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+
+# transform to training space
+def transform(x):
+    sigma_y = 0.10
+    return jnp.log(x + 1) / sigma_y
+# ------------------------------------------------- #
+
+
+# inheritance from Distribution class
+class NNPrior(dist.Distribution):
```
pmelchior, line 109: Let's be more specific here and call it `ScorePrior` because we might have neural network priors of different types in the future.

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+    support = constraints.real_vector
+
+    """Prior distribution based on a neural network"""
+
+    # construct custom vector-jacobian product
+
+    def __init__(self, trained_model='None', shape=1, validate_args=None):
```
pmelchior, line 116: `trained_model` -> `model`, which is the actual model or its file path. And let's store it in `self.model` rather than a global variable. Let's also store `self.shape`, but we should be able to get that from the model itself, right?

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+        # TODO: what's up with these globals?
+        # NOTE: this is a hack to get around the fact that    
+        # jax doesn't allow flexible arguments to be passed   
+        # to the log_prob function                           
+        global custom_model
+        custom_model = trained_model
+
+        # TODO: needs shape of the model images
+        event_shape = jnp.shape(shape)
```
pmelchior, line 125: `trained_model` -> `model`, which is the actual model or its file path. And let's store it in `self.model` rather than a global variable. Let's also store `self.shape`, but we should be able to get that from the model itself, right?

scarlet2/nn.py
```diff
@@ -0,0 +1,156 @@
...
+        nn_grad = vgrad(transform, x) * nn_grad # chain rule
+        return 0.0, nn_grad  # cannot directly call log_prob in Class object
+
+    def log_prob_bwd(res, g):
+        nn_grad = res # Get residuals computed in f_fwd
+        vjp = (g * nn_grad,) # create the vector (g) jacobian (nn_grad) product
+        return vjp
+    
+    # register the custom vjp 
+    log_prob.defvjp(log_prob_fwd, log_prob_bwd)
```
pmelchior, line 156: I think that can be done cleaner, i.e. without a global variable. See [here](https://github.com/google/jax/discussions/16540#discussioncomment-6265352).

I mean something like this (all before of the class context)
```
from functools import partial
@partial(jax.custom_vjp, nondiff_argnums=(0,))
def _log_prob(model, x):
    return 0
    
def log_prob_fwd(model, x):
    x = transform(x)
    nn_grad = calc_grad(x, model)
    ...
```
and then inside of the prior class, you should be able to do this:
```
def log_prob(self, x):
    return _log_prob(self, x)
```
It might even work to pull the partial trick _inside_ of the class definition, which means that you don't have to effectively duplicate the `log_prob` method.

scarlet2/nn.py
SampsonML: this has been altered as suggested in nn.py

scarlet2/nn.py
SampsonML: removed these redundancies, score model now specified by user at runtime.

scarlet2/nn.py
SampsonML: renamed

scarlet2/nn.py
SampsonML: renamed

scarlet2/nn.py
SampsonML: This is all done here (https://github.com/pmelchior/scarlet2/blob/matt-nn-branch/scarlet2/nn.py) will update pull request. Sample updated runscript here (https://github.com/pmelchior/scarlet2/blob/matt-nn-branch/sample_runscript.ipynb)

Note, the training data transform can now be specified by the user when initialising the prior then the Jacobian transform to convert to correct gradients is done with jax.

scarlet2/plot.py
```diff
@@ -0,0 +1,883 @@
...
+                img_to_rgb(model, norm=norm, channel_map=channel_map, mask=model_mask),
+                extent=extent,
+                origin="lower",
+            )
+            ax[k-skipped][panel].set_title("Model Source {}".format(k), fontsize=f_size)
+            if center is not None and add_markers:
+                ax[k-skipped][panel].plot(*center, "wx", **marker_kwargs)
+            panel += 1
+            
+        if show_hallucination:
```
pmelchior, line 633: does this work if the morphology does not have a prior? Also, does this assume that the prior is a score models?

SampsonML: Must have a prior, does not need to be a score model, just the log_prob() call must return the gradients for this method to work.

scarlet2/plot.py
```diff
@@ -0,0 +1,883 @@
...
+import jax
+import jax.numpy as jnp
+import jax.random as random
+from jax import jvp, grad, jit
+from astropy.visualization.lupton_rgb import LinearMapping, AsinhMapping
+from .bbox import Box
+
+
+#from .component import Component
+#from .source import NullSource
```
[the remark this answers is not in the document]
SampsonML: gone

scarlet2/plot.py
SampsonML: Added an assert statement asserting the morphology prior must no be `None`

pmelchior (approved):

PR #21, opened 2023-10-20
added convergence criteria

SampsonML:
Here, we can directly access the model updates from the optimiser which is equivalent to (x_old - x). This bypasses the need to look at gradients directly, which was what I was doing previously. If the L2 norm of the updates is less than the user-defined relative tolerance of the current data then we have converged.

### DIFF
--- a/scarlet2/scene.py
+++ b/scarlet2/scene.py
@@ -175,16 +175,16 @@ class Scene(Module):
         with tqdm.trange(max_iter, disable=not progress_bar) as t:
             for step in t:
                 # optimizer step
-                scene_, loss, opt_state, convergence = _make_step(scene, observations, optim, opt_state,
-                                                                  filter_spec=filter_spec,
-                                                                  constraint_fn=constraint_fn)
+                scene, loss, opt_state, convergence = _make_step(scene, observations, optim, opt_state,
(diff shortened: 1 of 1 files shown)


scarlet2/scene.py
```diff
@@ -175,23 +175,32 @@ def update_fn(updates, state, params):
...
-
-                # terminate optimization if all parameter change less than e_rel
+                
+                # terminate optimization if all gradient change less than e_rel
                 if e_rel is not None:
-                    crit = lambda x, x_: jnp.linalg.norm(x - x_) < e_rel * jnp.linalg.norm(x_)
-                    converged = tuple(
-                        crit(p, p_) for (p, p_) in zip(scene.parameters.values(), scene_.parameters.values()))
-                    if all(converged):
+                    crit = lambda x, x_: jnp.linalg.norm(x) < e_rel * jnp.linalg.norm(x_)
```
pmelchior, line 189: You are assuming a specific set of `Parameter` instances to be given (spectrum and morphology in this case), but this may not be the case (or some of them may be frozen). I suggest to do this crit check more generally, namely by using `jax.tree_util.tree_reduce`, so that you can go through all of the pytree leaves automatically.

pmelchior (approved):

PR #22, opened 2023-11-09
added confidence metric

SampsonML:
Hallucination score plotted during the scarlet2.plot.scene(show_hallucinations=True) call will now show -hallucination * src.morphology() where hallucination is defined as in Sampson/Melchior ICML

### DIFF
--- a/scarlet2/plot.py
+++ b/scarlet2/plot.py
@@ -8,8 +8,7 @@ import jax.numpy as jnp
 import jax.random as random
 from jax import jvp, grad, jit
 from .bbox import Box
-
-
+import matplotlib.colors as colors
 
 def channels_to_rgb(channels):
     """Get the linear mapping of multiple channels to RGB channels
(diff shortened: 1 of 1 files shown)


scarlet2/plot.py
```diff
@@ -632,18 +647,20 @@ def sources(
...
-                vmin=-true_max, 
-                vmax=true_max,
-                cmap="RdBu_r",
+                norm=colors.SymLogNorm(linthresh=1, linscale=1,
+                        vmin=-true_max, vmax=true_max, base=10),
+                cmap="RdBu",
                 extent=extent,
                 origin="lower",
             )
             colorbar = plt.colorbar(im, ax=ax[k-skipped][panel])
```
pmelchior, line 660: Please remove the colorbar: it's not particularly useful and creates different image dimensions for the model and the hallucination images.

scarlet2/plot.py
```diff
@@ -632,18 +647,20 @@ def sources(
...
-            ax[k-skipped][panel].set_title("Hallucination Source {}".format(k), fontsize=f_size)
+            colorbar.set_label('hallucination score', rotation=270, labelpad=20, fontsize=f_size)
+            title = 'confidence: ' + str(jnp.round(metric,3))
```
pmelchior, line 662: In keeping with the other labels: uppercase "Confidence", please

scarlet2/plot.py
SampsonML: Gone

scarlet2/plot.py
SampsonML: Updated

pmelchior (approved):

PR #23, opened 2023-12-04
changed default temperature from 0.0 to 0.02

SampsonML:
Empirically, having a temperature not identically 0 works a bit nicer. This could be an option for the user to set, as I previously had but this is very temperamental so I think the user should only be able to change it if they are really sure.

### DIFF
--- a/scarlet2/nn.py
+++ b/scarlet2/nn.py
@@ -96,16 +96,16 @@ def calc_grad(x, model, model_size=32):
     -------
     score_func : array of the score function
     """
-    # perform padding if needed
+    t = 0.02 # temperature: here we set a fixed value which should only be changed if the user is familiar with the model
     x = jnp.float32(x) # cast to float32
     x, pad_lo, pad_hi, pad = pad_fwd(x, model_size)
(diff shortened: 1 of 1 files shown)


scarlet2/nn.py
```diff
@@ -96,16 +96,16 @@ def calc_grad(x, model, model_size=32):
     -------
     score_func : array of the score function
     """
-    # perform padding if needed
+    t = 0.02 # temperature: here we set a fixed value which should only be changed if the user is familiar with the model
```
pmelchior, line 99: I dislike this implicit setting. The users should indeed by able to set this themselves, but than can simply be achieved by creating a `__call__` function in the model specification that knows of the temperature. E.g. 
```
class ModelWithTemp:
  def __init__(self, model, temp):
    ...
  def __call__(self, x):
    return self.model(x, t=self.temp)
```

pmelchior:
I think this should be solved differently, see above. Suggest to close without merging.

SampsonML:
Will refactor this, closing request

PR #25, opened 2023-12-07
altered routine slightly allowing for temp

SampsonML:
Model now compatible for a user define temperature to be sure during model construction via something akin to  'prior = nn.ScorePrior(model=prior_model, transform=transform, model_size=64, temperature=0.1)' Defaults are set inside the model call, instead of the score network architecture for more modular code.


SampsonML:
Fair, altered to mirror comment above, tested, working well.

pmelchior (approved):

PR #29, opened 2023-12-12
New version of iteration dependence

jaredcsiegel:
Iteration counting inside of scene is now replaced by having optax schedule as input parameter to scene.fit. Ex:

schedule = optax.cosine_onecycle_schedule(
    transition_steps=maxiter, 
    peak_value=1, 
    pct_start=0.5, 
    div_factor=10, 
    final_div_factor=.1)

scene.fit(obs, schedule, max_iter=max_iter, e_rel=e_rel) 

if schedule = None (default) no schedule is used.

### DIFF
--- a/scarlet2/scene.py
+++ b/scarlet2/scene.py
@@ -110,7 +110,7 @@ class Scene(Module):
         mcmc.run(rng_key, self, obs=observations)
         return mcmc
 
-    def fit(self, observations, max_iter=100, e_rel=1e-4, progress_bar=True, callback=None, **kwargs):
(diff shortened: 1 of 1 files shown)


pmelchior:
I like it. Did you check that it works?

pmelchior (approved):

jaredcsiegel:
Yes, I've verified that changing the scheduler has the expected effect on the fits (such as setting the scheduler to very large steps at late iterations) and that setting the scheduler=None gives identical results to manually turning off the scheduler in scene.py

PR #31, opened 2024-01-26
row/col swapped correctly plot.py

SampsonML:
Fixed minor issue in box cut for visualisations


pmelchior:
funny, very obvious that this is the right way now, not sure how the wrong way got into the code

pmelchior (approved):

Issue #34, opened 2024-02-13
correct behavior when constraint and prior are set


pmelchior:
It's easy to get into trouble when a parameter specifies both constraint and prior. The problem stems from a transformation that is done to the variable to make it appear (for the sampler or the optimizer) to be unconstrained. E.g a positive constraint amounts to adding $f(x) = \exp(x)$ into the path, which means that the original variable is reparameterized as $y=\log(x)$. For the likelihood, this transformation is transparent because autograd will just apply one more chain rule, but for a prior (e.g. a score model), we compute the gradient $\nabla \log p(x)$, not $\nabla \log p(y)$. This will get people into trouble without even knowing it.

So, we need a warning for those cases that the same transformation $f$ needs to be apply when training and testing the prior network.

SampsonML:
Aye, perhaps we can start writing some docs? Happy to get a start on that to fill up some busy work time.

jaredcsiegel:
[Draft (based on my current understanding of the issue, so this summary will likely be subject to edits)]

Yesterday, Peter and I further discussed the nuances of using both constants and priors. The story is a little different from the initial description from Peter above. In short, the gradients are evaluated in terms of the untransformed variables $x$ but are then applied to the transformed variables $y$.

For a little summary:

When a constraint is applied to the parameter x, it preforms the transformation: $x \rightarrow y = f(x)$ (for example, $f=\log$ if a positive constraint is applied).

The transformation is initially performed when $\texttt{scene.fit}$ is called:
"""
# transform to unconstrained parameters
constraint_fn = {name: biject_to(info["constraint"]) for name, (value, info) in parameters.items() if info["constraint"] is not None}
scene = _constraint_replace(self, constraint_fn, inv=True)
"""
at this point $x \rightarrow y$ for all parameters with a constraint.

The gradients for optimization are evaluating inside $\texttt{_make_step}$:
"""
loss, grads = eqx.filter_value_and_grad(loss_fn)(model)
"""
at this point $x \rightarrow y$ for all parameters with a constraint inside $\texttt{model}$.

Inside $\texttt{_make_step.loss_fn}$ the first step is to transform $y \rightarrow x$:
"""
model = _constraint_replace(model, constraint_fn)
"""
the likelihood and prior are then evaluated on $x$.

This creates a mismatch, because the gradients returned by $\texttt{loss_fn}$ are in terms of $x$ but these gradients are then applied to $\texttt{model}$, which is in terms of $y$:
"""
[...]

PR #35, opened 2024-02-17
Parameters with priors and constraints

pmelchior:
Following the discussion in #34, this PR brings a correction to the treatment of `ScorePrior`. In particular, there is no `ScorePrior.transform` anymore because these transformations are already applied _before_ the prior gets to see the parameters. Autodiff then takes care of the Jacobians, so we simply don't need to track them.

In addition, I streamlined the padding code. And the score model implementation can now also deal with arbitrary dimensions, so we can specify priors on SED, etc.

### DIFF
--- a/scarlet2/nn.py
+++ b/scarlet2/nn.py
@@ -15,73 +15,63 @@ import jax.numpy as jnp
 from jax import custom_vjp
 from jax import vjp
 
-
-
-def pad_fwd(x, model_size=32):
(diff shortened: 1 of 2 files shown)


SampsonML (approved):
Very nice, this is a much cleaner implementation. All modifications are now made to the trained galaxygrad package. This takes away any potential user error issues with transformations/constraints.

PR #36, opened 2024-02-20
Animation / free mpl styling args

pmelchior:
This PR brings mpl animations to `plot.scene`, which is the most relevant figure to see the evolution of the optimization or sampling procedure.

If called with a single `Scene`, it behaves as is. But if it get a list of scenes, it will turn in into an animation, where the images that depend on the state of the model (the model image itself, rendered, and residuals) will be updated. To get the list of scene from the fitter, use the `callback` argument:

```python
scenes = []
def scene_saver(scene, *args):
    scenes.append(scene)

scene_ = scene.fit(obs, max_iter=200, e_rel=0, callback=scene_saver)
ani = plot.scene(scenes, obs, show_model=True, show_rendered=True, show_observed=True, show_residual=True)
```

The output can be displayed inline in jupyter or saved to file, e.g. this gem with a terrible initialization:

https://github.com/pmelchior/scarlet2/assets/1463403/0b2df65f-ba2b-48cc-bbb6-6c8136da67e8

This PR also replaces hard-coded styling parameters, like size of the title fonts or colors of the source markers, with kwargs, so that they can be overwritten by the user:

```python
plot.observation(obs, 
    add_labels=False, 
    show_psf=True, 
    fig_kwargs={"figsize": (6,4)}, 
    title_kwargs={"color": "red", "size": 22}
)
```

### DIFF
--- a/scarlet2/plot.py
+++ b/scarlet2/plot.py
@@ -1,14 +1,17 @@
 from abc import ABC, abstractmethod
-import numpy as np
-import matplotlib.pyplot as plt
-from matplotlib.patches import Rectangle, Polygon
-from matplotlib.ticker import MaxNLocator
+
 import jax
 import jax.numpy as jnp
 import jax.random as random
-from jax import jvp, grad, jit
-from .bbox import Box
+import matplotlib.animation as animation
 import matplotlib.colors as colors
+import matplotlib.pyplot as plt
+import numpy as np
+from jax import jvp, grad, jit
+from matplotlib.patches import Rectangle, Polygon
+
+from .bbox import Box
+
 
 def channels_to_rgb(channels):
     """Get the linear mapping of multiple channels to RGB channels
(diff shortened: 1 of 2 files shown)


pmelchior:
I should add that I had some trouble with the animation showing in PyCharm, but that seems to be a problem of the IDE. And for saving, the best quality and smallest file sizes are from the `ffmpeg` writer, 
```python
ani.save("test.mp4", writer="ffmpeg")
```
which you have to install separately:
```
conda install -c conda-forge ffmpeg
```

SampsonML:
The scene returned in callback needs to run through the constraints otherwise colours/morph incorrect.  By adding 

```
                # report current iteration results to callback
                if callback is not None:
                    if constraint_fn is not None:
                        scene_frame = _constraint_replace(scene, constraint_fn)
                    else:
                        scene_frame = scene
                    callback(scene_frame, convergence, loss)
```

this is solved, in a relatively straightforward way


https://github.com/pmelchior/scarlet2/assets/44228257/d417f08e-3ef4-4c79-b644-dc8cafc44b02

SampsonML:
Also, this should work on most (maybe all) IDEs for visuals
```
from IPython.display import HTML
%matplotlib notebook
# create the animation
ani = plot.scene(scenes, obs, norm=norm, show_model=True, show_rendered=True, show_observed=True, show_residual=True)
# Show the animation
HTML(ani.to_html5_video())
```

pmelchior:
Good catch with the constraints. I should have seen this myself, with the colors going wild in the example above. Repeated now with the change: 

https://github.com/pmelchior/scarlet2/assets/1463403/3d7e7247-714a-4050-a343-39e817f010ad

Still very bad initialization (just 2 sources, one with uniform random morphology...), but that looks like a fit now...

SampsonML (approved):
Nice, in-built animations, very cool for presentations.

PR #43, opened 2024-03-14
Lanczos resampling renderer

b-remy:
This PR enables resampling a model frame to an observation frame with a Lanczos kernel. More precisely, here is what is added from the current main version:
- [x] channel renderer (collapse channels that are not needed)
- [x] Modular deconvolution/resampling/convolution renderers
- [x] `interpolation.py`, with lanczcos image resampling
- [x] 4 times image padding (as advised in [Bernstein & Gruen (2014)](https://arxiv.org/abs/1401.2636))
- [x] switch (if it's the same  `pixel_size`, use the already existing `ConvolutionRenderer`)

In addition, I have added
- [x] a `setup.py` for pip install
- [x] frame generator from observations ([from scarlet](https://github.com/pmelchior/scarlet/blob/45187fdccbe8b8c8d7551a84572d7243b24bc8cb/scarlet/frame.py#L156), this was not needed but useful for my tests)

For a quick demo of HST to HSC resampling, you can run this notebook https://colab.research.google.com/drive/1aA-ynB0BZiD7hL2PhrbHFyvxzRCC7CgY?usp=sharing

Let me know your thought and I you want me to add anything!

I am leaving this PR as a draft today because I couldn't figure out a way to jit `obs.render` and wanted to get your feedback to enable the feature. Have you tried this before @pmelchior @SampsonML @charlotteaward  @jaredcsiegel ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ setup(
     packages=['scarlet2'],
     install_requires=[
         'equinox',
+        'jax',
         'astropy',
         'numpy',
         'matplotlib'
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,18 @@
+from setuptools import setup
+
+setup(
+    name='scarlet2',
+    version='0.1.0',    
+    description='Scarlet, all new and shiny',
+    url='https://github.com/pmelchior/scarlet2',
+    author='Peter Melchior',
+    author_email='melchior@astro.princeton.edu',
+    license='MIT',
+    packages=['scarlet2'],
+    install_requires=[
+        'equinox',
+        'astropy',
+        'numpy',
+        'matplotlib'
+    ],
+)
--- a/scarlet2/fft.py
+++ b/scarlet2/fft.py
@@ -34,7 +34,7 @@ def transform(image, fft_shape, axes=None):
 
     image = _pad(image, fft_shape, axes)
     image = jnp.fft.ifftshift(image, axes)
-    image_fft = jnp.fft.fftn(image, axes=axes)
+    image_fft = jnp.fft.rfftn(image, axes=axes)
     return image_fft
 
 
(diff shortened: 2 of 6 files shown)


b-remy:
Just figured out jitting functions using equinox modules can be done with `equinox.filter_jit` and edited the colab notebook. I'm opening this PR for review :-)

setup.py
```diff
@@ -0,0 +1,18 @@
...
+    name='scarlet2',
+    version='0.1.0',    
+    description='Scarlet, all new and shiny',
+    url='https://github.com/pmelchior/scarlet2',
+    author='Peter Melchior',
+    author_email='melchior@astro.princeton.edu',
+    license='MIT',
+    packages=['scarlet2'],
+    install_requires=[
+        'equinox',
```
pmelchior, line 13: we should probably add jax here...

pmelchior:
Something occurred to me when going through the colab notebook: `Frame.bbox` needs to be an instance of `Box`, i.e. it is just bit of metadata to explain the shape and origin of the data array held by an Observation or by our internal model. We're using boxes to define the placement of source models or data because we may get e.g. a WCS for a large image, but the `Observation` instance will only hold a small portion. So, we define the bbox with a non-zero `origin` instead of constructing a new WCS.

I've seen you set `Frame.bbox` up to hold the actual data, but that's what `Observation.data` is for. Please change the notebook accordingly and also check that this misunderstanding doesn't affect the scarlet code.

b-remy:
Thank you @pmelchior . Indeed, I updated the notebook with
```python
hst_resampled = obs_hsc.render(obs_hst.data)
```
and it works fine as well. I just changed the `frame.from_observations` method where the frames bbox are `Box` now instead or arrays. I also added jax as a requirement in the package setup.py. In addition, I also implemented the galsim approximation for `x<1e-4` in the Lancsoz kernel.

PR #45, opened 2024-03-24
Adding Sersic morphology model

charlotteaward:
Adding Jared's Sersic2DMorphology as a new option for source morphology

### DIFF
--- a/scarlet2/__init__.py
+++ b/scarlet2/__init__.py
@@ -2,7 +2,7 @@ from . import plot
 from .bbox import Box
 from .frame import Frame
 from .module import Parameter, Module, relative_step
-from .morphology import Morphology, ArrayMorphology, GaussianMorphology, Sersic2DMorphology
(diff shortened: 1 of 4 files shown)


pmelchior:
For reference here: https://github.com/pmelchior/scarlet2/pull/47#issuecomment-2032792187

The relevant code in the scarlet1 repo is here:
https://github.com/pmelchior/scarlet/blob/master/scarlet/morphology.py#L210-L281

If you want, I can take the first stab on porting it to scarlet2, and then you can reimplement the (elliptical) Gaussian  we need for #47, the Sersic and the Spergel profile models.

pmelchior:
@charlotteaward I've coded up the profile morphology and the Gaussian and Sersic. Can you follow the approach here and bring over the SpergelProfile from scarlet1?

charlotteaward:
Fantastic, thank you! I'll work on the SpergelProfile.

pmelchior (approved):

Issue #33, opened 2024-02-01
In built initializations


SampsonML:
Functions for Scarlet2 default initialisations for the morphology and spectrum are included here: https://github.com/pmelchior/scarlet2/blob/initializations/scarlet2/initialization.py

To run may simply modify your fit routine as so adding two lines

```
init_spec = scarlet2.initialization.init_spectrum(obs, center)
init_morph = scarlet2.initialization.init_morphology(obs, center)
```
So he new routine looks like,

```
with Scene(model_frame) as scene:
    for center in centers:
        # initialize the prior
        prior = nn.ScorePrior(
            model=prior_model, transform=transform, model_size=model_size
        )
        # initialize the spctrum and morphology with Scarlet2
        init_spec = scarlet2.initialization.init_spectrum(obs, center)
        init_morph = scarlet2.initialization.init_morphology(obs, center)
        
        # construct the source
        Source(
            center,
            ArraySpectrum(
                Parameter(init_spec, constraint=constraints.positive, stepsize=5e-2)
            ),
            ArrayMorphology(Parameter(init_morph, prior=prior, stepsize=5e-3)),
        )

# now fit the model
scene_fit = scene.fit(obs, max_iter=220, e_rel=1e-4)  
renders = obs.render(scene_fit())
```

Currently, fitting for the boxsize is not done very nicely, so brainstorming and comments welcome. This is also the speed bottleneck in the morphology initialisations too.

Overall this works well producing some reasonable initialisations especially given our more robust optimisation scheme.
![inits](https://github.com/pmelchior/scarlet2/assets/44228257/74ae0209-a605-4e83-8873-78285365d6ff)
![renders](https://github.com/pmelchior/scarlet2/assets/44228257/1b60456e-dc2e-4b35-9bf0-6b7f56215dec)

SampsonML:
Note, this adds dependencies for `dm_pix` and `jaxopt` both pip installable.

SampsonML:
@pmelchior I will change the box size routine based on the scarlet1 methods, can you link where this happens in the scarlet1 repo

pmelchior:
Sure. Here's the relevant bit from the initialization of `SingleExtendedSource`:
https://github.com/pmelchior/scarlet/blob/45187fdccbe8b8c8d7551a84572d7243b24bc8cb/scarlet/source.py#L425-L439

The main steps are:
[...]

PR #47, opened 2024-03-31
Automated initialisations

SampsonML:
Adding automated source initialisation for the spectrum and morphology. Functions for single component and multi-component sources, where multi-component morphologies are returned as a list.

Example running is as follows, note a center _must_ still be defined source each source.

`init_spec = initialization.init_spectrum(obs, center)
  init_morph = initialization.init_morphology(obs, center, noise_thresh=400, corr_thresh=0.5, components=2)`


scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+# ---------------------------------------------------------------------------- #
+
+import numpy as np
+import numpy.ma as ma
+import jax.numpy as jnp
+import jax
+from .plot import cut_square_box
+from .morphology import GaussianMorphology
+from jaxopt import ScipyBoundedMinimize
+import dm_pix
```
pmelchior, line 20: Another dependency I like to avoid. You using it to rotate a Gaussian, which can be done analytically. #45 will bring this functionality.

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+        # simplified spectrum calculation for perimeter average around cutout
+        models = (None,) * len(observation.data)
+        spectra = []
+        spectrum_avg = 0
+        for obs, model in zip((observation,), models):
+            for i in range(perimeter_pixels.shape[1]):
+                spectrum = perimeter_pixels[:, i]
+                psf_model = obs.frame.psf()
+                # psf_model = jnp.expand_dims(psf_model, axis=0)
+                psf_peak = psf_model.max(axis=(1, 2))
```
pmelchior, line 182: That should be outside of the `for i ... ` loop

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+    sigma_x : float standard deviation of the Gaussian in the x direction
+    sigma_y : float standard deviation of the Gaussian in the y direction
+
+    Returns
+    -------
+    jnp.ndarray 2D Gaussian profile
+    """
+    return jnp.exp(
+        -((x - mu_x) ** 2 / (2 * sigma_x**2) + (y - mu_y) ** 2 / (2 * sigma_y**2))
+    )
```
pmelchior, line 63: We have that implemented in `GaussianMorphology`, so you should use that instead.

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+    data_cut = (data - jnp.min(data)) / (jnp.max(data) - jnp.min(data))
+    central_gaussian = create_gaussian_array(bx, 1.75, 1.75, 0)
+    fn = lambda x: jnp.mean(
+        central_gaussian * (data_cut - create_gaussian_array(bx, x[0], x[1], x[2])) ** 2
+    )
+    x0 = jnp.array([1.5, 1.5, 0])
+    lbfgsb = ScipyBoundedMinimize(fun=fn, method="l-bfgs-b", jit=True)
+    lower_bounds = jnp.array([1, 1, jnp.pi / 2])
+    upper_bounds = jnp.array([7, 7, jnp.pi])
+    bounds = (lower_bounds, upper_bounds)
```
pmelchior, line 118: these initial values I can understand but the limits are entirely arbitrary. They may well not work in practice. It would be much better to replace this by a unconstrained optimization on log(s)

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+    best_params = lbfgsb.run(x0, bounds=bounds).params
+    return best_params
+
+
+# take the 3 moments of a @D gaussian and construct the covariance matrix to fit the shape
```
pmelchior, line 123: That's thing that could completely remove the need for an optimizer in this place. Is suggest to go that route as part of `init_simple_morph`

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+        morph = np.array(morph, dtype=np.float32)  # convert to numpy to make mutable
+        rows, cols = morph.shape
+        central_row = rows // 2
+        central_col = cols // 2
+        morph[central_row, central_col] = 1.1  # extra brightening central pixel
+        morph = (morph - np.min(morph)) / (np.max(morph) - np.min(morph))
+        if bx > 30 and components == 2:
+            morph2 = create_gaussian_array(
+                bx, 1.25, 1.25, 0
+            )  # create a second component as a gaussian blob
```
pmelchior, line 254: I don't understand the heuristics that are used here. Why is this factor 1.1 there, what justifies the criterion for having 2 components (it looks like it's a size argument, which is indirectly a SNR argument), and what sets the size of the secondary component?

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+    Parameters
+    ----------
+    center: tuple
+        Position in the observation
+    observations: instance or list of `~scarlet.Observation`
+        Observation to extract SED from.
+    correct_psf: bool
+        If PSF shape variations in the observations should be corrected.
+    models: instance or list of arrays
+        Rendered models for this source in every observation
```
pmelchior, line 284: I think provinding the model is a scarlet1 option we should not longer support. Suggest to remove in this function

pmelchior:
For reference, `init_simple_morph` implements the idea formulated here: https://github.com/pmelchior/scarlet2/issues/33#issuecomment-1953298946

As for the need for a fitting method, I think it can be avoided. A Gaussian is uniquely defined by its mean $\mu$ and second moments, in the 2D there are 3 of them:

$$
Q_{ij} = \int d^2x I(x) (x_i - \mu_i) (x_j - \mu_j)
$$

So, we should port the `measure.py` module from scarlet1, because it has these higher moments.

Up to possible coordinate convention issues, this is the covariance matrix of a Gaussian approximation to the source. As it's done it weak lensing,  one can measure size $T = \sqrt{Q_{11}+Q_{22}}$ and (complex) ellipticity 

$$
\epsilon = \frac{Q_{11} - Q_{22} + 2\mathrm{i}Q_{12}}{Q_{11} + Q_{12} + 2\sqrt{Q_{11}Q_{22} - Q_{12}^2}} 
$$

There's an alternative ellipticity definition $\chi$, which is identical up to the term $2\sqrt{Q_{11}...}$, which it omits. However, $\epsilon$ is better aligned with the ellipticity definition you'd get from model fitting.

The inversion of mapping ellipticity and size to a transformation of a circular profile is coming with #45.

The only reason to be careful is what happens when other sources creep into the box you're considering because they will strongly increase all 2nd moments, but we increase the box sizes iteratively and stop before this problem becomes too severe.

SampsonML:
Working on reformulation of this, will heavily alter initial PR.

scarlet2/initialization.py
SampsonML: Removed

scarlet2/initialization.py
SampsonML: Removed fitting routine

scarlet2/initialization.py
SampsonML: Removed fitting routine

scarlet2/initialization.py
SampsonML: Removed fitting routine

scarlet2/initialization.py
SampsonML: Fixed

scarlet2/initialization.py
SampsonML: Removed fitting routine

scarlet2/initialization.py
```diff
@@ -0,0 +1,331 @@
...
+# just take the pixel value of the center of the source for each band.         #
+# ---------------------------------------------------------------------------- #
+
+import numpy as np
+import numpy.ma as ma
+import jax.numpy as jnp
+import jax
+from .plot import cut_square_box
+from .morphology import GaussianMorphology
+from jaxopt import ScipyBoundedMinimize
```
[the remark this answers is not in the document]
SampsonML: Removed fitting routine

scarlet2/initialization.py
```diff
@@ -0,0 +1,304 @@
...
+    Returns
+    -------
+    morph: `jnp.array` --> 2D Gaussian morphology
+    """
+    sigma = psf_sigma
+    converged = False
+    peak_spec = init_spectrum(observation, center, correct_psf=True)
+    spectra_prev = peak_spec
+    noise_rms = 1 / np.sqrt(ma.masked_equal(observation.weights, 0))
+    ma.set_fill_value(noise_rms, np.inf)
```
b-remy, line 117: This breaks in the case where `observation.weights=None`. I think we should either document  that we should note use None weights, or take into account this case in the code

scarlet2/initialization.py
```diff
@@ -0,0 +1,304 @@
...
+    morph : np.ndarray morphology of the source normalised to (0,1)
+    """
+    morph = init_simple_morph(obs, center, psf_sigma, noise_thresh, corr_thresh)
+    if (morph.shape[0] <= max_size) or (components == 1):
+        return morph
+
+    # fit for a more complex morphology for bigger sources
+    else:
+        bx = morph.shape[0]
+        cutout_obs = cut_square_box(obs.data, center, bx)
```
b-remy, line 210: I am testing this code on HSC data and `cut_square_box` fails for multichannel observations as it assumes 2D inputs. Am I missing something here? Or should we update this function to handle multichannel observations?

b-remy:
I mentioned two use cases where the code breaks, in particular for initializing the morphology of multiband data. I can provide a notebook if needed.

There are also a few methods in `mesure.py` that are not used in the initialization routine: `map_pixel`, `flux`, `centroid`, `snr` and `deconvolve`. Do we want to include them now?

Apart from these comments, it looks good to me.

scarlet2/initialization.py
SampsonML: Screenshot 2024-05-03 at 9 20 47 AM


This should be fixed now

scarlet2/initialization.py
SampsonML: For now added assertion that weights are not none

SampsonML:
> I mentioned two use cases where the code breaks, in particular for initializing the morphology of multiband data. I can provide a notebook if needed.
> 
> There are also a few methods in `mesure.py` that are not used in the initialization routine: `map_pixel`, `flux`, `centroid`, `snr` and `deconvolve`. Do we want to include them now?
> 
> Apart from these comments, it looks good to me.

I think lets leave them in for now, as I will at a point work on adding a deconvolution routine using some of these functions.

b-remy:
Alright, it looks good to me

scarlet2/measure.py
```diff
@@ -0,0 +1,205 @@
...
+def max_pixel(component):
+    """Determine pixel with maximum value
+
+    Parameters
+    ----------
+    component: `scarlet.Component` or array
+        Component to analyze or its hyperspectral model
+    """
+    if hasattr(component, "get_model"):
+        model = component.get_model()
```
b-remy, line 16: Just making a note here, we need to fix this before merging

pmelchior:
Where do we stand on this PR. Is there a code example/notebook that demonstrates the use of the initialization?

charlotteaward:
When I run this branch I'm finding the following bug:

> ConcretizationTypeError                   Traceback (most recent call last)
> Cell In[24], line 8
>       5 print(flux)
>       7 scene.set_info('sources.'+str(indtransient)+'.morphology.center', fixed=False,stepsize = pos_step)#, prior=dist.Normal(scene.sources[indtransient].morphology.center, 0.5))
> ----> 8 scene_ = scene.fit(observations_sc2, max_iter=n_steps, e_rel=1e-6,schedule=schedulepos)
>       9 flux = scene_.sources[indtransient].spectrum.data
>      10 print(flux)
> 
> File ~/scarletmerged2/scarlet2/scarlet2/scene.py:180, in Scene.fit(self, observations, schedule, max_iter, e_rel, progress_bar, callback, **kwargs)
>     177 with tqdm.trange(max_iter, disable=not progress_bar) as t:
>     178     for step in t:
>     179         # optimizer step
> --> 180         scene, loss, opt_state, convergence = _make_step(scene, observations, optim, opt_state,
>     181                                                          filter_spec=filter_spec,
>     182                                                          constraint_fn=constraint_fn)
>     184         # compute max change across all non-fixed parameters for convergence test
>     185         max_change = jax.tree_util.tree_reduce(lambda a, b: max(a, b), convergence)
> 
> File ~/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_jit.py:103, in _JitWrapper.__call__(self, *args, **kwargs)
>     102 def __call__(self, /, *args, **kwargs):
> --> 103     return self._call(False, args, kwargs)
> 
> File ~/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_jit.py:99, in _JitWrapper._call(self, is_lower, args, kwargs)
>      97         out = self._cached(dynamic, static)
>      98 else:
> ---> 99     out = self._cached(dynamic, static)
>     100 return _postprocess(out)
> 
>     [... skipping hidden 12 frame]
> 
> File ~/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_jit.py:37, in _filter_jit_cache.<locals>.fun_wrapped(dynamic, static)
>      35 fun = hashable_combine(dynamic_fun, static_fun)
>      36 args, kwargs = hashable_combine(dynamic_spec, static_spec)
> ---> 37 out = fun(*args, **kwargs)
>      38 dynamic_out, static_out = partition(out, is_array)
>      39 return dynamic_out, Static(static_out)
> 
> File ~/scarletmerged2/scarlet2/scarlet2/scene.py:252, in _make_step(model, observations, optim, opt_state, filter_spec, constraint_fn)
>     249         return loss_fn(model)
>     251     diff_model, static_model = eqx.partition(model, filter_spec)
> --> 252     loss, grads = filtered_loss_fn(diff_model, static_model)
>     254 updates, opt_state = optim.update(grads, opt_state, model)
>     255 model_ = eqx.apply_updates(model, updates)
> 
> File ~/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_ad.py:59, in _ValueAndGradWrapper.__call__(self, x, *args, **kwargs)
>      56     return self._fun(_x, *_args, **_kwargs)
>      58 diff_x, nondiff_x = partition(x, is_inexact_array)
> ---> 59 return fun_value_and_grad(diff_x, nondiff_x, *args, **kwargs)
> 
>     [... skipping hidden 8 frame]
> 
> File ~/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_ad.py:56, in _ValueAndGradWrapper.__call__.<locals>.fun_value_and_grad(_diff_x, _nondiff_x, *_args, **_kwargs)
>      53 @ft.partial(jax.value_and_grad, has_aux=self._has_aux, **self._gradkwargs)
>      54 def fun_value_and_grad(_diff_x, _nondiff_x, *_args, **_kwargs):
>      55     _x = combine(_diff_x, _nondiff_x)
> ---> 56     return self._fun(_x, *_args, **_kwargs)
> 
> File ~/scarletmerged2/scarlet2/scarlet2/scene.py:249, in _make_step.<locals>.filtered_loss_fn(diff_model, static_model)
>     246 @eqx.filter_value_and_grad
>     247 def filtered_loss_fn(diff_model, static_model):
>     248     model = eqx.combine(diff_model, static_model)
> --> 249     return loss_fn(model)
> 
> File ~/scarletmerged2/scarlet2/scarlet2/scene.py:234, in _make_step.<locals>.loss_fn(model)
>     228 if constraint_fn is not None:
>     229     # parameters now obey constraints
>     230     # transformation happens in the grad path, so gradients are wrt to unconstrained variables
>     231     # likelihood and prior grads transparently apply the Jacobians of these transformations
>     232     model = _constraint_replace(model, constraint_fn)
> --> 234 pred = model()
>     235 parameters = model.get_parameters(return_info=True)
>     236 log_like = sum(obs.log_likelihood(pred) for obs in observations)
> 
> File ~/scarletmerged2/scarlet2/scarlet2/scene.py:26, in Scene.__call__(self)
>      24 model = jnp.zeros(self.frame.bbox.shape)
>      25 for source in self.sources:
> ---> 26     model_ = source()
>      28     # cut out region from model, add single source model
>      29     bbox, bbox_ = overlap_slices(self.frame.bbox, source.bbox, return_boxes=True)
> 
> File ~/scarletmerged2/scarlet2/scarlet2/source.py:87, in Source.__call__(self)
>      85 def __call__(self):
>      86     base = super()
> ---> 87     model = base.__call__()
>      88     for component, op in zip(self.components, self.component_ops):
>      89         model_ = component()
> 
> File ~/scarletmerged2/scarlet2/scarlet2/source.py:27, in Component.__call__(self)
>      25 def __call__(self):
>      26     # Boxed model
> ---> 27     return self.spectrum()[:, None, None] * self.morphology()[None, :, :]
> 
> File ~/scarletmerged2/scarlet2/scarlet2/morphology.py:105, in GaussianMorphology.__call__(self)
>     102 def __call__(self):
>     103 
>     104     # faster circular 2D Gaussian: instead of N^2 evaluations, use outer product of 2 1D Gaussian evals
> --> 105     if (self.ellipticity==jnp.zeros((2,))).all():
>     107         _Y = jnp.arange(self.bbox.shape[-2]) + self.bbox.origin[-2] - self.center[-2]
>     108         _X = jnp.arange(self.bbox.shape[-1]) + self.bbox.origin[-1] - self.center[-1]
> 
>     [... skipping hidden 1 frame]
> 
> File ~/.conda/envs/scarlet/lib/python3.10/site-packages/jax/_src/core.py:1370, in concretization_function_error.<locals>.error(self, arg)
>    1369 def error(self, arg):
> -> 1370   raise ConcretizationTypeError(arg, fname_context)
> 
> ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: traced array with shape bool[].
> The problem arose with the `bool` function. 
> The error occurred while tracing the function _make_step at /home/cw1074/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_jit.py:32 for jit. This value became a tracer due to JAX operations on these lines:
> 
>   operation a:f32[] = convert_element_type[new_dtype=float32 weak_type=False] b
>     from line /home/cw1074/scarletmerged2/scarlet2/scarlet2/morphology.py:105 (__call__)
>

b-remy:
Ah, this is due to
```python
if (self.ellipticity==jnp.zeros((2,))).all():
```
In the `GaussianMorphology.__call__` function that I added in #50 . It was comparing the ellipticity with `jnp.zeros((2,))` which causes tracing problems in a `if` condition. I didn't check tracing. 

I just changed is with
```python
if abs(self.ellipticity).sum() == 0:
```
Let me know if it solves the error @charlotteaward :-)

charlotteaward:
Thanks for tracking it down! Unfortunately this new line still seems to give a Concretization Error. I wonder any call to self.ellipticity in an if statement would cause such a thing?

```
File ~/scarletmerged2/scarlet2/scarlet2/morphology.py:106, in GaussianMorphology.__call__(self)
    103 def __call__(self):
    104 
    105     # faster circular 2D Gaussian: instead of N^2 evaluations, use outer product of 2 1D Gaussian evals
--> 106     if abs(self.ellipticity).sum() == 0:
    107         _Y = jnp.arange(self.bbox.shape[-2]) + self.bbox.origin[-2] - self.center[-2]
    108         _X = jnp.arange(self.bbox.shape[-1]) + self.bbox.origin[-1] - self.center[-1]

    [... skipping hidden 1 frame]

File ~/.conda/envs/scarlet/lib/python3.10/site-packages/jax/_src/core.py:1370, in concretization_function_error.<locals>.error(self, arg)
   1369 def error(self, arg):
-> 1370   raise ConcretizationTypeError(arg, fname_context)

ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: traced array with shape bool[].
The problem arose with the `bool` function. 
The error occurred while tracing the function _make_step at /home/cw1074/.conda/envs/scarlet/lib/python3.10/site-packages/equinox/_jit.py:32 for jit. This value became a tracer due to JAX operations on these lines:

  operation a:f32[] = convert_element_type[new_dtype=float32 weak_type=False] b
    from line /home/cw1074/scarletmerged2/scarlet2/scarlet2/morphology.py:106 (__call__)

```

pmelchior:
Yeah, that's because Benjamin removed the `ellipticity=None` option internally. With it, models where ellipticity is None have a different traced path than those with ellipticity, but neither has an `if` statement inside of their respective forward call. Let me take a stab.

charlotteaward:
@b-remy One more question stemming from using this branch: when we initialize sources for a set of observations (and associated model frame) with a wcs vs no input wcs, do the expected units of the source center positions change from just pixel units? I'm having an issue where I if I initialize a source with pixel coordinates for observations created without using the wcs argument, it works correctly (first image), but the position is wrong if those observations had a given wcs (second image, where the galaxy center is wrong by >30 pixels and where the initialization routine is also finding the wrong flux and morphology because the position is being interpreted incorrectly, I've multiplied the initialization flux by 100).  
image
image

charlotteaward:
One other request - if we want to apply the positive constraint on the morphology, we need to make sure the morphology initialization has values >0. But since a jax array is returned, it's annoying to make a copy and adjust the values to be nonzero. Should we add an option to tell the initialization routine about the need to fulfill the positive constraint?

pmelchior:
As for the WCS: yes. If the observation has a WCS, coordinates are expected in RA/Dec. We could (and maybe should) make this more flexible, but that's what we have now.

As for positivity, I also ran into that problem. The easiest way to deal with it is to add a tiny number (like 1e-6) to all pixels. I'm just not sure where to do it, or how to do it in a way a user would understand.

charlotteaward:
Ok, hmm. I'm definitely getting the wrong positions when I input the center in ra and dec in either degrees or radians (see below). Also, the flux and morphology initialization code still requires that the center argument be in integer pixel coordinates, I wonder if we should generalize that to match the source initialization!
image

pmelchior:
The second problem will be fixed with my commits. The first one I don't understand.

pmelchior:
Can all of you test the new initialization code on examples you have. I refactored Matt's code and made use of existing methods in scarlet2.

Initialization can now be done e.g. with:
```python
from scarlet2 import *

center = jnp.array(center)
morph = init.adaptive_gaussian_morph(obs, center) + 1e-6
Source(
    center,
    ArraySpectrum(Parameter(spectrum, constraint=constraints.positive, stepsize=relative_step)),
    ArrayMorphology(Parameter(morph, constraint=constraints.positive, stepsize=0.1))
)
```

Note the addition of `1e-6` to prevent the positivity constraint transformation to yield `nan`s. Not the most elegant, but it works.

There are other methods, in particular `init.compact_morphology()`, which gives the model image of a point source.

pmelchior:
Added the ever-popular methods `set_spectrum_to_match`, which solves linear system of component spectra given morphologies. This gets the initial spectrum before the fit much closer to the actual image. Call it with
```python
scene.set_spectra_to_match(obs)
```
once all sources are in `scene`.

charlotteaward:
I updated the branch to test out the new morphology initialization but now I get all nans with adaptive_gaussian_morph. Would that be because I'm giving it a single band image?

pmelchior:
@charlotteaward are you placing positivity constraints on the morphology? And, if so, have you added a small number to prevent the nans for the log-transform?

charlotteaward:
@pmelchior I'm getting the nans in the output of initialization.adaptive_gaussian_morph(obs, centerpix), so before applying any positivity constraints in a source morphology declaration or doing any fitting. It's simulated data so no nans in the original imaging data either.

pmelchior:
`center` should be in sky coordinates, and the observations needs to have weights that are non-zero everywhere.

pmelchior:
@charlotteaward have you been able to track down where the nans come from?

charlotteaward:
@pmelchior Yes just checked and correcting the coordinate units fixed it up! I don't for these simulations, but I do often have zero weights coming from image masks in when running this on real data - should I switch that to instead using very tiny weights to make sure the initializations work?

pmelchior:
@charlotteaward No, we need to be able to handle zero weights. Let's check the new code in actual observations, please.

charlotteaward:
@pmelchior I'm testing it on low S/N ZTF data now, and find for a field with 8 galaxies it initializes the galaxies in the middle of the image just fine, and others as all nans. The nans arise for galaxies where part of the box (even if just the very edge) goes past the edge of the image. The behavior is the same with and without zero weights, so that's not a problem.

charlotteaward:
One other issue I'm running into with this branch. I'm finding that I can no longer refit the centers of point sources by running this before scene.fit: 
`scene.set_info('sources.0.morphology.center', fixed=False,stepsize = pos_step)`

When I try to sample over the center parameter (for point sources) I also get this error, even when the parameters are definitely valid:
```
RuntimeError                              Traceback (most recent call last)
Cell In[24], line 5
      3 scene_.set_info(scene_.parameters.keys(), fixed=True) 
      4 scene_.set_info('sources.'+str(indtransient)+'.morphology.center', fixed=False, prior=dist.Normal(scene_.sources[indtransient].morphology.center, 0.0001))
----> 5 mcmc = scene_.sample(observations_sc2,num_warmup=100, num_samples=300) 
      6 q = mcmc.get_samples()['sources.'+str(indtransient)+'.morphology.center'] 
      7 q1 = np.percentile(q,[2.5,16,50,84,97.5],axis=0) 

File ~/scarletmultisource/scarlet2/scarlet2/scene.py:111, in Scene.sample(self, observations, num_warmup, num_samples, **kwargs)
    109 mcmc = MCMC(nuts_kernel, num_warmup=num_warmup, num_samples=num_samples)
    110 rng_key = jax.random.PRNGKey(0)
--> 111 mcmc.run(rng_key, self, obs=observations)
    112 return mcmc

File ~/.conda/envs/scarlet/lib/python3.10/site-packages/numpyro/infer/mcmc.py:628, in MCMC.run(self, rng_key, extra_fields, init_params, *args, **kwargs)
    626 map_args = (rng_key, init_state, init_params)
    627 if self.num_chains == 1:
--> 628     states_flat, last_state = partial_map_fn(map_args)
    629     states = tree_map(lambda x: x[jnp.newaxis, ...], states_flat)
    630 else:

File ~/.conda/envs/scarlet/lib/python3.10/site-packages/numpyro/infer/mcmc.py:410, in MCMC._single_chain_mcmc(self, init, args, kwargs, collect_fields)
    408 # Check if _sample_fn is None, then we need to initialize the sampler.
    409 if init_state is None or (getattr(self.sampler, "_sample_fn", None) is None):
--> 410     new_init_state = self.sampler.init(
    411         rng_key,
    412         self.num_warmup,
    413         init_params,
    414         model_args=args,
    415         model_kwargs=kwargs,
    416     )
    417     init_state = new_init_state if init_state is None else init_state
    418 sample_fn, postprocess_fn = self._get_cached_fns()

File ~/.conda/envs/scarlet/lib/python3.10/site-packages/numpyro/infer/hmc.py:713, in HMC.init(self, rng_key, num_warmup, init_params, model_args, model_kwargs)
    708 # vectorized
    709 else:
    710     rng_key, rng_key_init_model = jnp.swapaxes(
    711         vmap(random.split)(rng_key), 0, 1
    712     )
--> 713 init_params = self._init_state(
    714     rng_key_init_model, model_args, model_kwargs, init_params
    715 )
    716 if self._potential_fn and init_params is None:
    717     raise ValueError(
    718         "Valid value of `init_params` must be provided with" " `potential_fn`."
    719     )

File ~/.conda/envs/scarlet/lib/python3.10/site-packages/numpyro/infer/hmc.py:657, in HMC._init_state(self, rng_key, model_args, model_kwargs, init_params)
    650 def _init_state(self, rng_key, model_args, model_kwargs, init_params):
    651     if self._model is not None:
    652         (
    653             new_init_params,
    654             potential_fn,
    655             postprocess_fn,
    656             model_trace,
--> 657         ) = initialize_model(
    658             rng_key,
    659             self._model,
    660             dynamic_args=True,
    661             init_strategy=self._init_strategy,
    662             model_args=model_args,
    663             model_kwargs=model_kwargs,
    664             forward_mode_differentiation=self._forward_mode_differentiation,
    665         )
    666         if init_params is None:
    667             init_params = new_init_params

File ~/.conda/envs/scarlet/lib/python3.10/site-packages/numpyro/infer/util.py:745, in initialize_model(rng_key, model, init_strategy, dynamic_args, model_args, model_kwargs, forward_mode_differentiation, validate_grad)
    732                             w.message.args = (
    733                                 "Site {}: {}".format(
    734                                     site["name"], w.message.args[0]
    735                                 ),
    736                             ) + w.message.args[1:]
    737                             warnings.showwarning(
    738                                 w.message,
    739                                 w.category,
   (...)
    743                                 line=w.line,
    744                             )
--> 745         raise RuntimeError(
    746             "Cannot find valid initial parameters. Please check your model again."
    747         )
    748 return ModelInfo(
    749     ParamInfo(init_params, pe, grad), potential_fn, postprocess_fn, model_trace
    750 )

RuntimeError: Cannot find valid initial parameters. Please check your model again.

```
This issue doesn't occur for other parameters like the spectrum.

pmelchior:
> Another interesting thing about these failing sources is that if I initialize them as point sources after the adaptive_gaussian_morph fails, and go ahead with the fitting, the loss functions goes to nans anyway

That is bizarre. Again, are you working with positivity constraints here?

For the point source center fitting, do you have other parameters that are being sampled? Are there any constraints?

charlotteaward:
@pmelchior I'm now trying both with and without positive constraints and see the same issues. For the point source center fitting, I'm only sampling the point source position and have fixed all other parameters, and get the error message with and without constraints. 

I tracked down why the log likelihood was also getting into nans when fitting those sources: the resampled ZTF PSF model images have some small negative values, and that was somehow producing nans in the log likelihood under particular conditions. I converted those small negative values to slightly positive values and the log likelihood was fixed, but it didn't solve the sampling error message or the nans in the initalizations.

pmelchior:
@charlotteaward you're finding a wide range of problems. Each of these should be treated as separate issue, so that we keep track of them without holding up this PR.

Let's get back to the original problem, which is that you get nans when using `init.adaptive_gaussian_morph`. Let's track this one down. A few questions:
1. Is the morphology directly after calling `adaptive_gaussian_morph` valid (i.e. no nans)?
2. Is the center coordinate within the observed image?
3. Is the spectrum from `pixel_spectrum` valid?
4. Is the spectrum valid after running `set_spectra_to_match(obs)`?

charlotteaward:
@pmelchior sounds good - I'm sure some issues stem from features of the ZTF data in particular, which manifest in subtle problems. To answer your questions, for the sources with nans coming out of the morphology initialization:

1. No, the nans are there in the array output of adaptive_gaussian_morph.
2. Yes. For 2 of 3 cases it's within 2 pixels of the edge, the third case is right in the center.
3. Yes.
4. Yes, still valid even for the sources with morphologies which are initialized as all nans.

The results are somewhat dependent on which single epoch image I choose for the initializations. Choosing a different image, I had only one edge source fail instead of 2 edge sources and one central source. The images are all pre-aligned with swarp.

pmelchior:
OK, so 1 is the immediate issue, but 4 I don't understand either. I was asking 3 because you wrote earlier that nans during optimization also happens when you initialize as point source. Can you check if the point source initialization produce nans on either spectrum or morphology, please

charlotteaward:
Can confirm that there are no nans in the point source spectrum or morphology!

pmelchior:
So, how did it happen that there were nans coming from fitting with the point source init?

charlotteaward:
That was because I was initializing those problematic galaxies with the morphology of the PSF model image, but that PSF model image had negative values in it, which caught me by surprise. It was then creating nans in the fitting because that violated the positive constraint.

pmelchior:
OK, this seems unrelated but still issue-worthy because that should not happen. As long as you don't fit a positivity constraint _on the PSF_ during optimization, the only effect would be that the rendered model assumes negative values, but that morphology should not have any problems staying positive.

pmelchior:
So, coming back to this PR. The problem appears to stem from the edge of the box being outside of the image. I'm betting that this is coming from `_get_edge_pixels(img, box)`. Can you check, please.

charlotteaward:
@pmelchior And T is coming out as nan in cases where ` (moments[(0, 2)] / flux * moments[(2, 0)] / flux - (moments[(1, 1)] / flux) ** 2)` in measure.size is negative, before we take it to the power of 1/4. So I guess something is wrong with the moments coming out of `measure.moments` in `gaussian_morphology`.

pmelchior:
Ok, that was bound to happen. These size and ellipticity measures are based
on moments, but we now have a weighted moment, where the weight function is
a step function at the edge of the image. We need some correction by
measuring the moments of that weight function. Check my DEIMOS paper for
low-order corrections

b-remy:
I am currently working on using these initialization routines for multiresolution data, it is working fine for both resolutions so far. Just a quick comment on what I have read above @pmelchior 
> center should be in sky coordinates, and the observations needs to have weights that are non-zero everywhere.

In my experience of the code I need to set the center in obs_frame pixel coordinates and I think it is what is expected in sources initialization:
```python
class Component(Module):
    spectrum: Spectrum
    morphology: Morphology

    def __init__(self, center, spectrum, morphology):
        self.spectrum = spectrum
        self.morphology = morphology
        self.morphology.center_bbox(center)
        super().__post_init__()
```
and
```python
class Morphology(Module):
    bbox: Box = eqx.field(static=True, init=False)

    def center_bbox(self, center):
        if isinstance(center, Parameter):
            center_ = center.value
        else:
            center_ = center
        self.bbox.set_center(center_.astype(int))
```

pmelchior:
@b-remy yeah, you're right, and it's part of #51.

That said, the thing to add for this PR is a bunch of checks in the init function if a user does something dodgy.
1. spectrum for a pixel that's not in the image
2. adaptive morphology for a source that's cut off at the edge
3. probably a few others...

I'm going to implement them as exceptions so that the user has to deal with them. Practically, I think the most straightforward approach is to not model a source that cannot be properly initialized. If these are all the edge of the postage stamp, that should not be too problematic. But in any case, we should alert the user when the problem occurs rather than waiting for it to blow up later.

pmelchior:
@charlotteaward I've added the exceptions. Can you check if these are triggered in the cases that gave you trouble, please.

charlotteaward:
@pmelchior yep it's triggering correctly! All looks good on my end.

pmelchior:
OK, good. @b-remy can you finish your review so that we can merge. There are a few issues being held up by this one. #48 would be one; and I have an idea for a better version of `scene.set_spectra_to_match`, which really doesn't work that well.

b-remy:
This PR has been extensively tested and it looks good to me to me, so you have my go to merge @pmelchior

b-remy (approved):

pmelchior:
Yay! Epic PR with basically everyone contributing!

PR #50, opened 2024-05-05
Profile morphology from ellipticities

b-remy:
Hi @SampsonML , @pmelchior ,

I think I found the issue with generating `ProfileMorphology` from ellipticities. 

The coordinate transformation matrix
$` \dfrac{1}{\sqrt{1-g²}} \begin{pmatrix} 1+g_1 & g_2 \\ g_2 & 1-g_1 \end{pmatrix}`$ (the covariance matrix is the inverse of this one), 

is defined with respect to $`g`$ which is the *reduced* definition of the ellipticity. However, the input parameters of the profile are $`e_1`$ and $`e_2`$ which are the *distortion* definition ([see here](https://galsim-developers.github.io/GalSim/_build/html/shear.html) for definitions and how to convert one from the other).

I am fine working with $`e_1`$ and $`e_2`$ and this is what is computed using the moments in [Melchior et al. (2010)](https://academic.oup.com/mnras/article/412/3/1552/1050244?login=false) eq. 2:

$\epsilon = [G_{20} - G_{02} + 2iG_{11}] / [G_{20} + G_{02}]$.

I am not familiar with the formula in #47 using an additional term in the denominator, and everything is consistent with the definition above.

Therefore, I think the `moments` method in `measure.py` is correct, we just need to make a conversion when using distortion ellipticities in `morphology.py` (hence this PR). For later, we may want to have the flexibility to take as input either distortion or reduced ellipticities, let me know if you want me to add these options.

[Here is a notebook](https://colab.research.google.com/drive/1cS6YnQNI0OyIPGIctQoLSQx33PG-ak4C?usp=sharing) which compares the two definitions (for ellipticity computation) and where the profile is generated using the conversion from distortion to reduced ellipticities.

### DIFF
--- a/scarlet2/morphology.py
+++ b/scarlet2/morphology.py
@@ -102,7 +102,7 @@ class GaussianMorphology(ProfileMorphology):
     def __call__(self):
 
         # faster circular 2D Gaussian: instead of N^2 evaluations, use outer product of 2 1D Gaussian evals
-        if self.ellipticity==jnp.zeros((2,)):
+        if (self.ellipticity==jnp.zeros((2,))).all():
 
             _Y = jnp.arange(self.bbox.shape[-2]) + self.bbox.origin[-2] - self.center[-2]
             _X = jnp.arange(self.bbox.shape[-1]) + self.bbox.origin[-1] - self.center[-1]
--- a/scarlet2/morphology.py
+++ b/scarlet2/morphology.py
@@ -37,19 +37,17 @@ class ProfileMorphology(Morphology):
     center: jnp.array
     size: float
     ellipticity: (None, jnp.array)
-    g: (None, jnp.array)
 
     def __init__(self, center, size, ellipticity=None, bbox=None):
 
         # define radial profile function
         self.center = center
         self.size = size
-        self.ellipticity = ellipticity
+        if ellipticity is None:
+            self.ellipticity = jnp.zeros((2,))
+        else:
+            self.ellipticity = ellipticity
 
-        g_factor = 1 / (1. + jnp.sqrt(1. - (ellipticity[0]**2 + ellipticity[1]**2)))
-        self.g = self.ellipticity * g_factor
-
-        self.ellipticity = ellipticity
         super().__post_init__()
 
         if bbox is None:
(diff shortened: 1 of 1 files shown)


SampsonML:
Nice,  from the notebook the size however is still overestimating by a bit bigger than a factor of root 2?

pmelchior:
OK, great. So, the take-aways: 
1. The elliptical profile definition is identical between `ProfileMorphology` and galsim's profiles. I might have looked at the galsim code when I created it years ago, not sure anymore..
2. Both use the simpler definition (which in Bartelmann & Schneider is called $\chi$ instead of $\epsilon$).
3. Both give moment sizes $T=47$ when we set it with a size of $\sigma=30$. 

3 is a result of 1, but nonetheless weird. One culprit can be the factor $1/\sqrt{1-e^2}$, which both we and galsim use in the profile definition. I tried setting $e=0$ and got a factor of $\sqrt{2}$ up to uncertainties from the pixel sizes, so this is still wrong. I wonder if this definition is meant to preserve flux or surface brightness or something else. Can one of you dig through the galsim docs to see if they mention it anywhere.

pmelchior:
OK, I found it. The correct form of the size given the flux-normalized 2nd moments of the profile is $det(Q)^{1/4}$, i.e.
```
(g[(0,2)] * g[(2,0)] - (g[(1,1)])**2)**(1/4)
```
Matt tried this earlier, but using only the exponent 1/2 instead of 1/4.

scarlet2/morphology.py
```diff
@@ -37,12 +37,18 @@ class ProfileMorphology(Morphology):
     center: jnp.array
     size: float
     ellipticity: (None, jnp.array)
+    g: (None, jnp.array)
```
pmelchior, line 40: Can you implement `g` as a class property instead of a class member, please. It's a quick computation, so we can avoid the redundancy of the definitions.

scarlet2/morphology.py
```diff
@@ -37,12 +37,18 @@ class ProfileMorphology(Morphology):
...
 
     def __init__(self, center, size, ellipticity=None, bbox=None):
 
         # define radial profile function
         self.center = center
         self.size = size
+        self.ellipticity = ellipticity
+
+        g_factor = 1 / (1. + jnp.sqrt(1. - (ellipticity[0]**2 + ellipticity[1]**2)))
```
pmelchior, line 49: You need to check if `ellipticity is not None` for this to work. Please add.

b-remy: OK, I added the check if ellipticity is None in the `__init__` and set it to `jnp.zeros((2,))` if so

scarlet2/morphology.py
b-remy: Sure, done.

pmelchior (approved):

PR #52, opened 2024-05-10
Update multiresolution renderer and test

b-remy:
This PR updates the resampling renderer strategy by interpolating all the images (observation and PSFs) on the same grid in Fourier space and adds k-wrapping ([Bernstein & Gruen 2014](https://arxiv.org/abs/1401.2636)) to match galsim resampling. Some refactoring has been done in `interpolation.py` as well and a test can be run with data soon [here](https://github.com/astro-data-lab/scarlet-test-data).

### DIFF
--- a/tests/test_multiresolution.py
+++ b/tests/test_multiresolution.py
@@ -36,7 +36,6 @@ psf_hst = np.array(psf_hst[None,:,:], np.float32)
 psf_hst = jnp.pad(psf_hst, ((0,0), (1,0), (1,0)))
 psf_hst = np.repeat(psf_hst, 5, 0)
 
-print(psf_hst.shape)
 psf_hst = scarlet2.ArrayPSF(psf_hst)
 
 # Scale the HST data
--- a/tests/test_multiresolution.py
+++ b/tests/test_multiresolution.py
@@ -1,9 +1,15 @@
+import os
 import warnings
 warnings.filterwarnings('ignore')
 
 import numpy as np
 from numpy.testing import assert_allclose
+
 import scarlet2
+
+from utils import import_scarlet_test_data
+import_scarlet_test_data()
+from scarlet_test_data import data_path, tests_path
 
 import numpy as np
 import astropy.io.fits as fits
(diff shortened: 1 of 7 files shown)


tests/test_multiresolution.py
```diff
@@ -0,0 +1,80 @@
...
+import scarlet2
+
+import numpy as np
+import astropy.io.fits as fits
+from astropy.wcs import WCS
+
+import jax.numpy as jnp
+
+# Load the HSC image data
+obs_hdu = fits.open('../scarlet-test-data/data/test_resampling/Cut_HSC1.fits')
```
pmelchior, line 15: Can we pull this file off github without getting the repo in this particular directory? FITS is technically a text-base file format, so what happens if we try to get the raw file with a URL request?

pmelchior:
This looks all good. I assume that the test runs well compared to galsim, so the last question is whether we can get rid of the relative path for loading those files. One could request that anyone running test needs to clone the test data repo, but that still doesn't produce an unambiguous file path. Hence my question if we can download files like that from github over the internet by a URL request.

But thinking about this some more, an alternative would be that `scarlet-test-repo` defines these file paths relative to _its_ installed location, so that one could this e.g. this:
```python
import scarlet_test_data.data_path as data_path
import astropy.io.fits as fits

obs_hdu = fits.open(sdata.resampling.obs_hsc_path)
```

Or possible even simpler:
```python
from scarlet_test_data import data_path
import astropy.io.fits as fits
import os

obs_hdu = fits.open(os.path.join(data_path, "test_resampling", "Cut_HSC1.fits"))
```

I like the last option best because it avoids defining a special variable simply to resolve a path where only the base directory is unknown.

The things that are therefore needed:
1. a `__init__.py` file for `scarlet-test-data` that checks the modules installed location, adds "/data" to it, and store that in the variable `data_path`.
2. our test scripts to start with a call to `pip install git+https://github.com/astro-data-lab/scarlet-test-data.git`

For 1, this should do:
```python
data_path = os.path.join(os.path.split(__file__)[0], "data")
```

I also think we need to define the package name with `_` instead of `-` because the latter cannot be used for imports.

b-remy:
I see your point and I agree it's much better without hard coded relative data paths. The way I fixed it is starting test files with
```python
from utils import import_scarlet_test_data
import_scarlet_test_data()
```
which pip installs the data repo if not already installed, and applied your `data_path` approach. Let me know if this works for you!

pmelchior (approved):
Good work! I really like the pip install

PR #56, opened 2024-05-22
New Parameter specification

pmelchior:
_this PR follows from #55_

The new interface splits creating and initializing the model from specifying optimization/sampling parameter. An example is  below:
```python

    for center in centers:
        # all inputs to models are int, float, complex or jnp arrays
        center = jnp.array(center)
        spectrum = 100 * jnp.ones(5)
        morph = jnp.ones((21, 21))

        Source(
            center,
            ArraySpectrum(spectrum),
            ArrayMorphology(morph)
        )

# create a lookup structure how to find a specific node of the equinox model scene so that it can be optimized
# can find variable names automatically, but better to give names explicitly
parameters = scene.make_parameters()
for i in range(len(scene.sources)):
    parameters += Parameter(scene.sources[i].spectrum.data, name=f"spectrum.{i}", constraint=constraints.positive, stepsize=spec_step)
    parameters += Parameter(scene.sources[i].morphology.data, name=f"morph.{i}", constraint=constraints.positive, stepsize=0.1)

# can remove parameters by name
parameters -= "spectrum.4"
```


scarlet2/scene.py
```diff
@@ -4,7 +4,7 @@
 
 from .bbox import overlap_slices
 from .frame import Frame
-from .module import Module, rgetattr
+from .module import Module, Parameters
```
b-remy, line 7: `Parameters` is not used in this file

b-remy (approved):
This looks good to me @pmelchior , there are a few conflict to solve though as the main branch has been updated in the meantime

PR #59, opened 2024-05-29
Minimum example as a test script

pmelchior:
Documents and test use on the quickstart example from the scarlet1 docs. Should be run to check if we have broken important part of the user-facing API

### DIFF
--- a/tests/test_multiresolution.py
+++ b/tests/test_multiresolution.py
@@ -2,16 +2,13 @@ import os
 import warnings
 warnings.filterwarnings('ignore')
 
-import numpy as np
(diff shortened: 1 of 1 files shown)


b-remy:
This looks good to me. I can only see changes in [tests/test_multiresolution.py](https://github.com/pmelchior/scarlet2/pull/59/files#diff-3f389a28404f1e1da5f5ad5e858c310949a790241125b8d0bec9ef1c9d6a5ced), but you are mentioning the quickstart example from scarlet1. Do you intend to add something else?

b-remy (approved):
Alright! Let's merge

PR #60, opened 2024-06-03
Multiresolution init and fitting demo

b-remy:
This PR enables multiresolution initialization and fitting. It also adds codes from scarlet1.wavelets to reproduce the [multiresolution tutorial](https://pmelchior.github.io/scarlet/tutorials/multiresolution.html) from scarlet1.

Here is the equivalent tutorial in scarlet2: https://colab.research.google.com/drive/1mvJN17r5traikPFrBICQTdhpxxaF1rLS?usp=sharing

I wonder if we should upload it on this repo at some point for documentation, or as a script similar to the quickstart test. It could also be in the scarlet-test-data repo. Let me know if you want me to push it somewhere and add some text to document what is happening :-)


scarlet2/observation.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, data, weights, psf=None, wcs=None, channels=None, renderer=No
     def C(self):
         return self.frame.C
 
+    def set_weights(self, weights):
```
pmelchior, line 39: I don't think this method is needed. Simply specify the weights at init time. They are not supposed to change afterwards anyway (this was just a little sloppy in the old docs)

scarlet2/wavelets.py
```diff
@@ -0,0 +1,181 @@
+# from https://github.com/pmelchior/scarlet/blob/master/scarlet/wavelet.py
+
+import jax.numpy as jnp
+import numpy as np
```
pmelchior, line 4: We should be able to do everything with jnp. That would remove any ambiguity on whether these are jax numpy variables or not.

scarlet2/observation.py
pmelchior: OK, I see why this was done in the docs. We simply didn't have the background rms for these observations without running `sep`. I'd still prefer to remove this method and instead rewrite the method `makeCatalog` from the notebook to accept the images (aka `obs.data`) instead.

scarlet2/wavelets.py
b-remy: Sure, this is done

pmelchior:
This works as intended. There are a few details to improve on in the notebook (in future PRs)

1. There's a confusing color shift in `plot.scene` if the filter sets are different because the mapping from the model channels to the observations cannot use the same channel map and norm. Probably applying the observation dependent `ChannelRenderer` when showing the pre-rendered model is the way to solve this.
2. The adaptive initialization in single-band HST cannot rely in changes in the color to limit the box size (we use this in multi-band as an indicator for a neighboring source). This may be the reason why some of the initial boxes are clearly too small (in particular source for 10).
3. And we need morphology priors for this example!

PR #64, opened 2024-06-05
starlet morphology

pmelchior:
This PR adds the starlet parameterization for source morphologies.

It's enabled by the starlet implementation from #60, but I needed to add the method `starlet_reconstruction` from scarlet1.

One can now do this:
```python
center = jnp.array(center)
morph = init.adaptive_gaussian_morph(obs, center, min_corr=0.99)
morph = StarletMorphology.from_image(morph)
Source(center, ArraySpectrum(spectrum), morph)
```

This parameterization is useful for spatially correlated structures (like low SFB features), but without additional constraints that model isn't terribly useful because the starlet basis is overcomplete. That is, if an image can describe the morphology, so can a starlet model. However, it get's interesting if these coefficients are constrained to be positive and sparse (only a few of them are non-zero). Sparsity is usually created by the loss function (L0 or L1 penalty on the coefficients), which is not something I would like to do. But in scarlet1, we used proximal operators to create the sparse solution. This could be done explicitly in `StarletMorphology.__call__`. I think one can stick the [soft thresholding operator](https://github.com/pmelchior/proxmin/blob/master/proxmin/operators.py#L138-L150) (which promotes L1 sparsity) into the forward path and still differentiate. I'll check.

### DIFF
--- a/scarlet2/__init__.py
+++ b/scarlet2/__init__.py
@@ -9,7 +9,7 @@ from . import plot
 from .bbox import Box
 from .frame import Frame
 from .module import Parameter, Module, relative_step
-from .morphology import Morphology, ArrayMorphology, GaussianMorphology, SersicMorphology
+from .morphology import Morphology, ArrayMorphology, GaussianMorphology, SersicMorphology, StarletMorphology
 from .observation import Observation
 from .psf import PSF, ArrayPSF, GaussianPSF
 from .scene import Scene
--- a/scarlet2/morphology.py
+++ b/scarlet2/morphology.py
@@ -4,6 +4,7 @@ import jax.scipy
 
 from .bbox import Box
 from .module import Module
+from .wavelets import starlet_transform, starlet_reconstruction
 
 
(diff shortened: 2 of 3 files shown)


pmelchior:
One can indeed add the proximal operators in the forward path, so I added these options:
```python
morph = StarletMorphology.from_image(morph, positive=False, l1_thresh=1e-2)
```
for the threshold level and whether positivity of the coefficients is required. Just as a note, this is a deviation of our usual approach in putting the constraints in the hand of the user, because we specify these one as part of the model. The reason is that our usual approach (namely to have constraints that are applied to parameters to transform into a unconstrained representation) doesn't work for them because the soft thresholding operation is not invertible. I think it's reasonable to add this modeling option here instead of being dogmatic about the types of models we can fit.

b-remy (approved):
This looks all good to me @pmelchior , as a bonus, here is the multiresolution notebook using StarletMorphology at initialization :-)

PR #65, opened 2024-06-05
Deimos

### LINKED ISSUE #48: DEIMOS for Gaussian initializations
The Gaussian morphology inits from #47 amount to determine the moments of the observed image distribution. That means the forward pass will convolved them with the PSF again. It would be better if we could initialize them at their preconvolved values. And it turns out this can be done (as I derived [here](https://doi.org/10.1111/j.1365-2966.2010.17875.x)). There's an analytic mapping between the moments of the convolved $\lbrace G^\star \rbrace$, the deconvolved $\lbrace G \rbrace$, and the PSF moments $\lbrace P \rbrace$, which I conveniently summarized in Table 1:

Screenshot 2024-04-05 at 16 26 57

In this the moments are defined as $\lbrace I \rbrace_{i,j} = \int d^2x I(x)x_1^i x_2^j$. So, these equations can be directly inverted analytically. Even better, I coded up the entire shebang in C++ _a long time ago_:
https://github.com/pmelchior/shapelens/blob/32393c390c8f8dada5448f120fbcd6d8ecb74e84/src/DEIMOS.cc#L89-L126

So, we should port this to python and add it to the `measure` module that will be brought from scarlet1 as part of #47.

pmelchior:
This PR implements the moment deconvolution method (closes #48), so that an initial Gaussian from `adaptive_gaussian_morph` is computed for the deconvolved model frame rather then the convolved observed frame.

It also updates the spectrum using the zeroth moment, which is the integrated flux. This is somewhat dangerous in the presence of nearby objects (in that case `pixel_spectrum` is the safest choice), but it has lower variance (more pixel contribute).

Because of the deconvolution, the moments might be ill-defined, which will throw a `ValueError`, so the recommended pattern is:
```python
try:
    spectrum, morph = init.adaptive_gaussian_morph(obs, center, min_corr=0.99)
except ValueError:
    morph = init.compact_morphology()
```

### DIFF
--- a/scarlet2/bbox.py
+++ b/scarlet2/bbox.py
@@ -112,6 +112,12 @@ class Box(eqx.Module):
         """
         return tuple([slice(o, o + s) for o, s in zip(self.origin, self.shape)])
 
+    @property
+    def spatial(self):
+        """Spatial component of higher-dimensional box"""
+        assert self.D >= 2
+        return self[-2:]
+
     def set_center(self, pos):
         """Center box at given position
         """
--- a/scarlet2/initialization.py
+++ b/scarlet2/initialization.py
@@ -53,13 +53,12 @@ def adaptive_gaussian_morph(obs, center, min_size=11, delta_size=3, min_snr=20, 
     last_spectrum = peak_spectrum.copy()
     box2d = Box((min_size, min_size))
     center_pix = obs.frame.get_pixel(center)
-    if not obs.frame.bbox[-2:].contains(center_pix):
+    if not obs.frame.bbox.spatial.contains(center_pix):
         raise ValueError(f"Pixel coordinate expected, got {center_pix}")
     box2d.set_center(center_pix.astype(int))
 
     # increase box size until SNR is below threshold or spectrum changes significantly
-    while max(box2d.shape) < max(obs.data.shape[-2:]):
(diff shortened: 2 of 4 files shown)


b-remy:
Yes since #56, `Parameters` are to be set after the scene initialization. This should work instead:
```python
from numpyro.distributions import constraints
from scarlet2 import relative_step
from functools import partial 

spec_step = partial(relative_step, factor=5e-3)
morph_step = partial(relative_step, factor=5e-3)

    for center in centers:

        try:
            spectrum, morph = init.adaptive_gaussian_morph(obs, center, min_corr=0.99)
        except ValueError:
            morph = init.compact_morphology()
            
        print(f"morph shape is {np.shape(morph)}")
        print(f"spectrum shape is {np.shape(spectrum)}")

        Source(
            center,
            ArraySpectrum(spectrum),
            ArrayMorphology(morph),
        )


parameters = scene.make_parameters()
for i in range(len(scene.sources)):
    parameters += Parameter(scene.sources[i].spectrum.data, name=f"spectrum.{i}", constraint=constraints.positive, stepsize=spec_step)
    parameters += Parameter(scene.sources[i].morphology.data, name=f"morph.{i}", prior=prior, stepsize=morph_step)
```

Issue #51, opened 2024-05-10
Consistent treatment of coordinates


pmelchior:
We make implicit assumptions about the coordinate frame, in particular for source centers. When no WCS is set in the observed frame, they _must_ be in pixel coordinates, but if a WCS is set, they _may_ be in pixel coordinates or in sky coordinates. We assume it's the latter, but that may not be true and lead to confusions, e.g. this:

_Originally posted by @charlotteaward in https://github.com/pmelchior/scarlet2/issues/47#issuecomment-2103497628_

As we already depend on `astropy` for the WCS transformation, it would make sense to allow positions to be given as `SkyCoord` (in which case it will be transformed according to a WCS) or a jnp.array (in which case we assume it to be in pixel coordinates).

charlotteaward:
Quick note that we also need to adjust plot.py, which assumes pixel coordinates for sources when adding labels!

pmelchior:
After more conversation about this, we should also make sure that all operations involving the user can be made in either all WCS or all pixel coordinates. This is particularly important for multi-observation cases.

I suggest that we define all model coordinates in the frame of the sky. That means that center coordinates would be in RA/Dec, with step sizes in, say, arcsec; same for sizes of radial profiles. We can make use of the `astropy.units` framework to automate conversions.

This will provide one important advantage: when a user defines the model frame, they can now simply say it's an abstract  piece of sky, with a center and a size. From that perspective, it's not an image that is tacked onto the sky, which has routinely led to difficulties with new users. We, internally, compute images, but they represent that sky, so it makes sense to assume the coordinates of the sky, not of an image of the sky.

For this to work, we need to allow for two modes:
* Everything is in pixels: A trivial WCS that implements the identity mapping will suffice for that.
* Everything is in sky: our method `Frame.from_observations` already supports that mode. I'd like to add a method that constructs a WCS from center, size, and pixel scale.

pmelchior:
To further clarify our approach, I think we should allow the _definition_ of model properties and parameters in sky coordinates, e.g.:
```python
import astropy.units as u

center = astropy.coordinates.SkyCoord(ra, dec)
size = 1*u.arcsec
morph = GaussianMorphology(center, size)

# using the parameter framework from PR #56 
parameters += Parameter(morph.center, stepsize=0.01*u.arcsec)
```

For fitting/sampling we now have two choices:
[...]

PR #66, opened 2024-06-10
Consistent coordinates

b-remy:
This PR solves #51 . The code will be able to accept either pixel coordinates and quantities as well as sky coordianates and astropy quantities. Here are some details of the implementation:

1. `frame.get_pixel` now expects `SkyCoord` objects (or list of) and returns pixel coordinates as arrays and `frame.get_sky_coord` expects pixel coordiantes as arrays and returns `SkyCoord` (or list of), to remove any ambiguity. If it's an array, then it is pixel coordinates, if it is `SkyCoord`, then it is sky coordinates.
2. Anytime a `SkyCoord` of `u.Quantity` object is used in `__init__`, it looks for the scene and convert it to pixels
3. Same thing for `Parameter`, but in addition:
a. I had to move the content of `set_constraint()` from the `Parameter.__init__` to `Parameter.__iadd__`because we need to convert the constraints parameters to pixels before checking the constraints, which expects arrays and floating arguments.
b. numpyro priors require a special treatement because we not only need to update the distribution parameters (e.g. loc and scale) but also the batch and event shape. For instance
```python
numpyro.dist.Normal(loc=SkyCoord(ra=0, dec=0, unit='deg'), scale=1*u.arcsec)
```
would set `batch_shape=0` and `event_shape=0` and converting loc and scale to pixels does not automatically update the batch_shape, while
```python
numpyro.dist.Normal(loc=[0., 0.], scale=1)
```
would set `batch_shape=2` and `event_shape=0`

[Here is a notebook](https://colab.research.google.com/drive/1ZzQHT0OgoGKOaRzu66yhD5sYlCYVFak8?usp=sharing) which uses `SkyCoord` and `u.Quantity` for all fieldname in ['node', 'constraint', 'prior', 'stepsize']

### DIFF
--- a/scarlet2/frame.py
+++ b/scarlet2/frame.py
@@ -3,9 +3,10 @@ import equinox as eqx
 import jax.numpy as jnp
 import numpy as np
 
+from astropy.coordinates import SkyCoord
+
 from .bbox import Box
 from .psf import PSF
-
 
 class Frame(eqx.Module):
     bbox: Box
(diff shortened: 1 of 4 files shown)


scarlet2/frame.py
```diff
@@ -38,75 +40,92 @@ def pixel_size(self):
...
 
-        if self.wcs is not None:
-            wcs_ = self.wcs.celestial  # only use celestial portion
-            pixel = jnp.array(wcs_.world_to_pixel_values(sky)).reshape(-1, 2)
-            # y/x instead of x/y:
-            pixel = jnp.flip(pixel, axis=-1)
+        if isinstance(sky_coord, list):
+            assert all(isinstance(coord, SkyCoord) for coord in sky_coord)
+            pixel = jnp.asarray([coord.to_pixel(wcs_) for coord in sky_coord],
+                                 dtype="float32")
```
pmelchior, line 60: I don't think this is necessary. `SkyCoord` works also with lists of coordinates and should also transform lists of coordinates: https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html

scarlet2/frame.py
```diff
@@ -38,75 +40,92 @@ def pixel_size(self):
...
+        astropy.coordinates.SkyCoord
+        """
+        pixels = pixels.reshape(-1, 2)
 
-        if sky.size == 2:
-            return sky[0]
-        return sky
+        assert self.wcs is not None
+        wcs = self.wcs.celestial # only use celestial portion
+        sky_coord = [SkyCoord.from_pixel(pixel[0], pixel[1], wcs) for pixel in pixels]
```
pmelchior, line 83: Same here, `SkyCoord` should transform ndarrays in one go: https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.from_pixel

scarlet2/frame.py
```diff
@@ -38,75 +40,92 @@ def pixel_size(self):
...
+            -------
+            size in pixels
+        """
+        assert u.get_physical_type(size) == "angle"
+        # first computer the pixel size
+        pixel_size = get_pixel_size(
+            get_affine(self.wcs.celestial) # only use celestial portion
+        ) # in deg/pixel
+        
+        return size.to(u.deg).value / pixel_size
```
pmelchior, line 128: I know it's arbitrary, but degrees is an inconvenient unit for pixel sizes. I'd choose arcsec instead if this doesn't cause any trouble elsewhere

b-remy:
Simplified `frame.get_sky_coord` as suggested and changed `frame.get_pixel` accordingly. Also updated the conversion now performed in arcsec. The tests in the notebook still pass!

pmelchior (approved):
This looks good. I made a minor change to convert centers coordinates for PointSources as well.

b-remy:
Right, thank you @pmelchior !

Issue #73, opened 2024-07-10
multi-observation inits


pmelchior:
None of the init methods were tested for multi-observation scenarios. The main reason is that it's not clear at all how that should be done. For instance, `spectrum = init.pixel_spectrum(observations, center)` can imply a number of things that are supposed to happen:
1. `observations` contain different channels, and the resulting `spectrum` gets a concatenation in the order of observations. However, this may not be the order of channels in the model frame. In this mode, we should read the model frame channels and associate the observed channel with the model channel accordingly.
2. `observations` contain (at least some) identical channels. Should we average the results? Maybe.

So, we could improve this by implementing a channel check as in item 1 and maybe a warning in case 2.

The situation gets more complicated for `adaptive_morphology`. The method finds the largest bounding box around `center` to contain light that has the same color as the center pixel. It then measures the moments of the image in that box to create a Gaussian with those moments.  If the placement or resolution of the observations is different, we have to make adjustments to how this Gaussian moments are combined. In particular, we can take advantage of DEIMOS (#65), which means we have deconvolved moments, so the effects of different PSFs are already taken out. What's missing is that different resolutions means a different numbers of pixel for an objects with the same physical size. But that correction is also analytic: the correct factor is $(s_\mathrm{frame} / s_\mathrm{obs})^p$ for the $p$-th moment. So, we can combine the moments across different observations.

_Originally posted by @pmelchior in https://github.com/pmelchior/scarlet2/issues/72#issuecomment-2221127686_

pmelchior:
The pixel spectrum is straightforward for multi-observations, but the adaptive morphology is trickier. What we basically attempt to do is to determine the moments of the observed images, but in the frame of the model. That has two different steps:
1. Determine the size of the box in the observations that contain (most of) the source but little else (background, neighbors). It's not clear initially if the boxes should be the same across different observations or be allowed to adjust independently. The second one is more modular, but the first one may be more robust, especially to detect the presence of neighbors.
[...]

PR #74, opened 2024-07-16
Multiobs initialization and moments conversion

b-remy:
Hi @pmelchior ,

I'm opening this PR to make the code review for #73  easier. I added on my side the code for converting moments from one WCS to another. I also added a few tests targeting rescaling, rotation, flip and arbitrary WCS conversion, including the case where the observations are mono or multichannels. You can also take a look at [this notebook](https://colab.research.google.com/drive/1BspxcnCr2wmIn_HSuYRve2mttovhRU19?usp=sharing) which illustrates these tests :-)

### DIFF
--- a/scarlet2/initialization.py
+++ b/scarlet2/initialization.py
@@ -1,3 +1,6 @@
+import operator
+from functools import reduce
+
 import jax.numpy as jnp
 from astropy.coordinates import SkyCoord
 
@@ -240,6 +243,37 @@ def pixel_spectrum(obs, pos, correct_psf=False, return_array=False):
     -------
     spectrum: `~jnp.array`, ArraySpectrum, or list thereof if given list of observations
     """
+
+    # for multiple observations, get spectrum from each observation and then combine channels in order of model frame
+    if isinstance(obs, (list, tuple)):
+
+        # flat lists of spectra and channels in order of observations
+        spectra = jnp.concatenate(
+            [pixel_spectrum(obs_, pos, correct_psf=correct_psf, return_array=True) for obs_ in obs])
+        channels = reduce(operator.add, [obs_.frame.channels for obs_ in obs])
+
+        try:
+            frame = Scenery.scene.frame
+        except AttributeError:
+            print("Multi-observation initialization can only be created within the context of a Scene")
+            print("Use 'with Scene(frame) as scene: ...")
+            raise
+
+        spectrum = []
+        for channel in frame.channels:
+            try:
+                idx = channels.index(channel)
+                spectrum.append(spectra[idx])
+            except ValueError:
+                msg = f"Channel '{channel}' not found in observations. Setting amplitude to 0."
+                print(msg)
+                spectrum.append(0)
+        spectrum = jnp.array(spectrum)
+
+        if return_array:
+            return spectrum
+        return ArraySpectrum(spectrum)
+
     assert isinstance(obs, Observation)
 
     if isinstance(pos, SkyCoord):
--- a/scarlet2/initialization.py
+++ b/scarlet2/initialization.py
@@ -2,7 +2,6 @@ import operator
 from functools import reduce
 
 import jax.numpy as jnp
-from astropy.coordinates import SkyCoord
 
 from . import Scenery
 from . import measure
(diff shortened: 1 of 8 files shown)


scarlet2/measure.py
```diff
@@ -124,71 +126,215 @@ def moments(component, N=2, center=None, weight=None):
...
+                    center = jnp.asarray(center)
+                    # centroid wrt given center
+                    self[1, 0] -= center[0] * self[0, 0]
+                    self[0, 1] -= center[1] * self[0, 0]
+                if model.ndim == 3 and center[0].ndim == 1:
+                    center = center[0][:, None, None], center[1][:, None, None]
+                grid_y = grid_y - center[0]
+                grid_x = grid_x - center[1]
+    
+    def higher_order_moments(self):
```
pmelchior, line 188: What is this function for? From the facto that it goes from moment N to 2N, I'm guessing to get the variances...

However, the constructor can measure any moment you want, but this code requires the model and the weights to be saved, which I am not a fan of because there's ample change to modify one without the others updating.

tests/test_quickstart.py
```diff
@@ -29,7 +29,7 @@ def test_quickstart():
         for center in centers:
             center = jnp.array(center)
             try:
-                spectrum, morph = init.adaptive_morphology(obs, center, min_corr=0.99)
+                spectrum, morph = init.make_bbox(obs, center, min_corr=0.99)
```
pmelchior, line 32: This must be incorrect. That method only returns a box, but we need `spectrum, morph = init.adaptive_morphology(obs, center, min_corr=0.99)`. Please change back.

pmelchior:
Overall, this looks very good. Happy to standardize moments. This will be useful for good inits, provided we there's no garbage in the image, like neighboring galaxies ;-)

scarlet2/measure.py
b-remy: In order to update moments from 0 to N, we also need to use moments from N+1 to 2N (see [Teague (1980), eq. 36](https://opg.optica.org/josa/fulltext.cfm?uri=josa-70-8-920&id=57703)), so I proposed to compute these additional moments in another function to update the first N moments stored the class. I am not a fan either that it requires to have access to the model. The problem is that once we updated the first N moments, the moments up to 2N are outdated since still correspond to the untransformed original image, so cannot be used for another transformation... 

Unless we keep track of the transformations in the class, and if the object is affected by another transformation we know the corresponding higher moments of the original image. What do you think?

pmelchior: I don't see what you mean. Teague's eq. 36 clearly shows that you don't need higher moments. The paper even clarifies this point in the text below.

scarlet2/initialization.py
```diff
@@ -23,9 +25,9 @@ def _get_edge_pixels(img, box):
     return jnp.concatenate(edge, axis=1)
 
 
-def adaptive_morphology(obs, center, min_size=11, delta_size=3, min_snr=20, min_corr=0.99, return_array=False):
+def make_bbox(obs, center_pix, min_size=11, delta_size=3, min_snr=20, min_corr=0.99):
     """
-    Create image of a Gaussian from the centered 2nd moments of the observation.
+    Make a bounding box for source at center.
```
b-remy, line 30: This function, which used to return the adaptive morphology, now only returns the box. I agree we should revert the change, but I wanted to double check with you if this was intended. I'm happy to make the change back.

tests/test_quickstart.py
b-remy: Right, I agree! Please check the related comment bellow

pmelchior: OK, we're both wrong. The drop-in replacement for this method is `init.from_gaussian_moments`. Please change that

scarlet2/initialization.py
pmelchior: Yes, that is fine. I made the code more modular, so this function was broken out. No changes needed here.

scarlet2/measure.py
pmelchior: This still is an open issue. I would much prefer the `Moment` instances not holding on to the data they were computed with.

tests/test_quickstart.py
b-remy: Right, this is fixed now :-)

scarlet2/measure.py
b-remy: I got confused due to the j+k-r-s index, which I thought could go up to 4 if j and k are both equal to 2. But the sum is never above 2 for 2nd order moments (j=m, k=n-m)

b-remy: This should be fixed now :-)

pmelchior (approved):
all good. go ahead and merge

Issue #75, opened 2024-07-17
WCS in resampling renderers


b-remy:
The current implementation of resampling renderers does not take into account a change in WCS between observation and model frames. Scarlet was doing it and I assume that Scarlet2 should as well. We have all the pieces to implement this (especially since #74 ), so I am on it!

PR #76, opened 2024-08-14
Change of WCS in multiresolution renderer

b-remy:
This PR solves #75. So far the multiresolution renderer only worked for aligned WCSs with a scale difference. Now, the code detects WCSs rotations in `obs.match()` and handles them in `obs.render()`.

### DIFF
--- a/scarlet2/frame.py
+++ b/scarlet2/frame.py
@@ -7,6 +7,7 @@ from astropy.coordinates import SkyCoord
 
 from .bbox import Box
 from .psf import PSF
+from .measure import get_scale
 
 
 class Frame(eqx.Module):
@@ -35,7 +36,8 @@ class Frame(eqx.Module):
     @property
     def pixel_size(self):
         if self.wcs is not None:
-            return get_pixel_size(get_affine(self.wcs)) * 60 * 60  # in arcsec
+            # return get_pixel_size(get_affine(self.wcs)) * 60 * 60  # in arcsec
+            return get_scale(self.wcs.wcs).mean() * 60 * 60 # in arcsec
         else:
             return 1
 
--- a/scarlet2/frame.py
+++ b/scarlet2/frame.py
@@ -7,8 +7,6 @@ from astropy.coordinates import SkyCoord
 
 from .bbox import Box
 from .psf import PSF
-from .measure import get_scale
-
 
 class Frame(eqx.Module):
     bbox: Box
@@ -37,7 +35,7 @@ class Frame(eqx.Module):
     def pixel_size(self):
         if self.wcs is not None:
             # return get_pixel_size(get_affine(self.wcs)) * 60 * 60  # in arcsec
-            return get_scale(self.wcs.wcs).mean() * 60 * 60 # in arcsec
+            return get_scale(self.wcs).mean() * 60 * 60 # in arcsec
         else:
             return 1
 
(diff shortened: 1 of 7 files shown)


scarlet2/frame.py
```diff
@@ -35,7 +36,8 @@ def C(self):
     @property
     def pixel_size(self):
         if self.wcs is not None:
-            return get_pixel_size(get_affine(self.wcs)) * 60 * 60  # in arcsec
+            # return get_pixel_size(get_affine(self.wcs)) * 60 * 60  # in arcsec
+            return get_scale(self.wcs.wcs).mean() * 60 * 60 # in arcsec
```
pmelchior, line 40: moving the scale computation to `measure` is not my preferred choice. `measure` should be things we measure from scarlet models, these functions refer to sizes and angles from WCS. I suggest moving them to `frame.py` instead.

Also, the call `get_affine` is to catch different WCS formatting conventions. See https://github.com/pmelchior/scarlet/pull/286 for that issue in scarlet1.

scarlet2/interpolation.py
```diff
@@ -266,18 +266,38 @@ def body_fun_y(i, args):
...
+    if phi is not None:
+        R = jnp.array([[jnp.cos(phi), jnp.sin(phi)],
+                       [-jnp.sin(phi), jnp.cos(phi)]])
+        
+        b_shape = kcoords_out.shape
+        kcoords_out = (R @ kcoords_out.reshape((-1, 2)).T).T.reshape((b_shape))
+
+    # TODO: Apply flip on rotation if any
+    # if flip_sign is not None:
+    #     kcoords_out = kcoords_out*flip_sign.reshape((2,1,1))
```
pmelchior, line 299: Do we still need this sign flip piece. If not, please remove. But also see the next comment

scarlet2/interpolation.py
```diff
@@ -286,6 +306,8 @@ def resample_ops(kimage, shape_in, shape_out, res_in, res_out, interpolant=Quint
         interpolant
         )
     
+    # TODO: Apply flip on rotation if any
+    # k_resampled *= flip_sign.prod()
```
pmelchior, line 310: If I'm not completely mistaken, you need the `flip_sign` at some place. Otherwise, please remove from method signature.

scarlet2/renderer.py
```diff
@@ -179,15 +181,32 @@ def __init__(self, model_frame, obs_frame, padding=None):
         object.__setattr__(self, "res_in", model_frame.pixel_size)
         object.__setattr__(self, "res_out", obs_frame.pixel_size)
 
+        # Extract rotation angle between WCSs using jacobian matrices
+        angle_in = get_angle(model_frame.wcs.wcs)
+        angle_out = get_angle(obs_frame.wcs.wcs)
```
pmelchior, line 186: same problem as above. Use `get_affine` instead.

pmelchior:
Where do we stand on this PR?

b-remy:
I moved `get_scale`, `get_angle` and `get_sign` to `frame.py` which is indeed cleaner and updated the code accordingly. 

I am not sure yet how to handle different sign conventions in the WCS pc, but since it is not so usual to have different sign conventions I am raising an error now. I think this PR can be merged and we can think about how to define resampling coordinates when there is a WCS flip later.

pmelchior (approved):
good to merge

PR #83, opened 2024-10-22
HDF5 output save

SampsonML:
This adds a routine for the saving and loading of a full Scarlet2 scene saves via pickling to a binary, and storing the binary as as HDF5 group attribute. This has been tested and runs returning the same Scene object.

`model_to_h5(filename, scene, ID, path, overwrite)` will save the scene object
`model_from_h5(filename, ID, path)` will load the object


scarlet2/__init__.py
```diff
@@ -16,3 +16,4 @@ class Scenery:
 from .source import Component, DustComponent, Source, PointSource
 from .spectrum import Spectrum, ArraySpectrum, StaticArraySpectrum, TransientArraySpectrum
 from .wavelets import Starlet
+from .io import model_to_h5, model_from_h5
```
pmelchior, line 19: Please remove this line. I don't want to force people to install hdf5 if they don't need this form of serialization

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
+import h5py
+import os
+import numpy as np
+import jax.numpy as jnp
+from scarlet2 import *
```
pmelchior, line 5: Add `from scarlet2.io import model_to_h5, model_from_h5`,

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
...
+import pickle
+import jax
+
+
+def model_to_h5(filename, scene, ID, path="", overwrite=True):
+    """
+    Save the model output to a single HDF5 file.
+    NOTE: This is not a pure function hence cannot be
+    utalised within a JAX JIT compilation.
```
pmelchior, line 14: -> utilized

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    # create directory if it does not exist
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+    # first serialize the model into a pytree
+    model_group = "scene_id_" + str(ID)
+    save_path = os.path.join(path, model_group)
+    save_h5_path = os.path.join(path, filename)
+    print(f"Saving model to {save_path}")
```
pmelchior, line 35: Remove line 33 and 35

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns
+    -------
+    None
+    """
+    # create directory if it does not exist
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+    # first serialize the model into a pytree
+    model_group = "scene_id_" + str(ID)
```
pmelchior, line 32: just `str(ID)`, it's more flexible.

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns
+    -------
+    scene : scarlet2.scene instance
+    """
+
+    filename = os.path.join(path, filename)
+    f = h5py.File(f"{filename}.h5", "r")
+    model_group = "scene_id_" + str(ID)
+    if model_group not in f.keys():
+        raise ValueError("ID not found in the file.")
```
pmelchior, line 75: `f"ID {ID} not found in file."`

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+
+    # save the binary to HDF5
+    group = f.create_group(model_group)
+    model = pickle.dumps(scene)
+    group.attrs["model"] = np.void(model)
+    f.close()
+
+
+    return None
```
pmelchior, line 54: not needed

scarlet2/io.py
```diff
@@ -0,0 +1,83 @@
+import h5py
+import os
+import numpy as np
+import jax.numpy as jnp
+from scarlet2 import *
+import pickle
+import jax
+
+
+def model_to_h5(filename, scene, ID, path="", overwrite=True):
```
pmelchior, line 10: change to default to `overwrite=False`, please

scarlet2/io.py
SampsonML: We need to explicitly close the H5 file to save right? return statement not needed agreed

SampsonML:
Changes made

scarlet2/io.py
SampsonML: In the test_save_outputs.py?

tests/test_save_output.py
pmelchior, line 1: I'd remove the plotting routines entirely. Just end with a comparison between the scene before and after the save/load calls.

SampsonML: Is there a standard way to do this in python, since even if the contents of `scene` are indentical, I cannot run `scene_save == scene_load` I believe since loading from a different file python will consider them different objects.

The scenes match when I compare pre-pickled and post-pickled inside the same file (ie inside io.py)

pmelchior: I think you can do it with `jax.tree_util` by comparing the tree structure and the each of the leafs

SampsonML: cool done and passing

tests/test_save_output.py
```diff
@@ -96,19 +84,11 @@ def test_save_output():
...
-        show_rendered=True,
-        show_observed=True,
-        show_residual=True,
-    )
-    plt.show()
 
+    # compare scenes 
+    saved = jax.tree_util.tree_structure(scene_)
+    loaded = jax.tree_util.tree_structure(scene_loaded)
+    print(f"saved == loaded: {saved == loaded}")
```
pmelchior, line 91: Can you add the test for `jax.tree_util.tree_leaves`, please. It probably complains about missing `__eq__`, in which case I'd add something like `for leave1, leave2 in zip(leaves1, leaves2): assert jnp.allclose(leave1, leave2)`.

tests/test_save_output.py
SampsonML: all set

tests/test_save_output.py
SampsonML: check updated now

tests/test_save_output.py
```diff
@@ -86,9 +86,14 @@ def test_save_output():
...
-    loaded = jax.tree_util.tree_structure(scene_loaded)
-    print(f"saved == loaded: {saved == loaded}")
+    saved = jax.tree_util.tree_leaves(scene_)
+    loaded = jax.tree_util.tree_leaves(scene_loaded)
+    status = True
+    for leaf_saved, leaf_loaded in zip(saved, loaded):
+        if (leaf_saved != leaf_loaded).all():
+            status = False
+        
+    print(f"saved == loaded: {status}")
```
pmelchior, line 96: One last thing: to work as a test, it needs to be pass clean or raise an error, so use `assert status is True, "Loaded leaves not identical to original"` instead of  / in addition to  `print`

pmelchior (approved):
Make sure to "squash and merge" to avoid bloating the repo history with unintended bulk like the h5 file you saved and then removed.

Issue #87, opened 2024-11-04
test_quickstart.py 


Majoburo:
# Bug Report: `ValueError` in `test_quickstart.py` Due to Unexpected `None` Handling in JAX (Workaround Available) 

## Description
Running `test_quickstart.py` results in a `ValueError: Expected None, got Array([ 5., 34.], dtype=float32)`. This error occurs during a call to `scene.fit()` and is likely related to recent changes in JAX's handling of `None` values in tree structures.

## Steps to Reproduce
1. Set up a clean Conda environment.
2. Install `scarlet2`, `NumPyro`, `Optax`, and other dependencies.
3. Run the `test_quickstart.py` script.

Example commands:
```bash
# Commands to reproduce the issue
conda create -n scarlet2 python
conda activate scarlet2
git clone https://github.com/pmelchior/scarlet2
cd scarlet2
pip install numpyro optax .
python tests/test_quickstart.py
```

## Environment
- **JAX version:** 0.4.35
- **JAXlib version:** 0.4.35

## Error Traceback
```python
python test_quickstart.py
...
ValueError: Expected None, got Array([ 5., 34.], dtype=float32).

In previous releases of JAX, flatten-up-to used to consider None to be a tree-prefix of non-None values. To obtain the previous behavior, you can usually write:
  jax.tree.map(lambda x, y: None if x is None else f(x, y), a, b, is_leaf=lambda x: x is None)
```

## Workaround
Forcing `jax` and `jaxlib` to version `0.4.28` resolves the issue and allows `test_quickstart.py` to run successfully.

pmelchior:
Thanks, Majo. so, somewhere between jax 0.4.30 and 0.4.35, some change was made the breaks our code. Can you post the entire bracktrace, so that we can see which call is causing the problem, please.

Majoburo:
Sure!

```python
python tests/test_quickstart.py
/Users/majoburo/Work/scarlet2/tests/test_quickstart.py:22: UserWarning: A JAX array is being set as static! This can result in unexpected behavior and is usually a mistake to do.
  obs = Observation(data, weights, psf=ArrayPSF(jnp.asarray(psf))).match(model_frame)
Traceback (most recent call last):
  File "/Users/majoburo/Work/scarlet2/tests/test_quickstart.py", line 66, in <module>
    test_quickstart()
    ~~~~~~~~~~~~~~~^^
[...]

PR #92, opened 2024-11-14
check for None in optax update

pmelchior:
This commit should address the problem, reported in #87, that was introduced by a change in jax 0.4.31 and another one in optax a while back.


pmelchior:
Tested on python3.10 with jax 0.4.26 and optax 0.2.2 and on python3.13 with jax 0.4.34 and optax 0.2.3

PR #118, opened 2025-04-08
removed TempScore class

SampsonML:
There is no reason for the child class TempScore. Temperature is directly added as a default, but adjustable parameter in ScorePrior.


SampsonML:
Ah, no my mistake - changed required still

pmelchior (approved):
All good. I just cleaned up the docstrings,

PR #119, opened 2025-04-08
How to model sources with multi-resolution observations

b-remy:
- New *How to model sources with multi-resolution observations* tutorial
- Also fixes `plot.scene` issue when scene and observation frames are not the same
  - Before  
image

  - After fixing labels and boxes coordinates
image

  - After fixing scene extent in observation frame
image


b-remy:
Updated the notebook with a model frame generated using the `Frame.from_observation`. The example above was using the HST frame as the model frame.

- `Frame.from_observation` was broken for `coverage="union"`, now fixed
- added a `AdjustToFrame` renderer which fits the rendered model into the observation box (by cropping and padding automatically)
- reintroduced `renderer: (Renderer, eqx.nn.Sequential) = eqx.field(static=True)` in `Observation` attributes (otherwise cannot jit `_make_step`)

Also removed the cells output. Here is the output of the uploaded tutorial (using `coverage=union`)

multi-res-union

b-remy:
The merge of #132 into main generated some issues with the initialization with multiple observations, the last commits propose to solve this by:
- setting the default argument of `box_sizes` in arcsec. I chose those values:
```python
box_sizes = [1.848*u.arcsec,
             2.856*u.arcsec,
             4.200*u.arcsec,
             5.880*u.arcsec,
             7.896*u.arcsec,
             10.24*u.arcsec,
             12.93*u.arcsec]
```

which correspond to the equivalent of the previous box sizes in pixels with the HSC resolution
- converting these angles in pixels for each observations and performing the same operations as before
- boxes sizes are then converted back in pixels in the scene frame for averaging
- Both pixel and angle `box_sizes` are accepted as argument of `from_gaussian_moment`
- But if the frame has no WCS (e.g. prior tutorial), then the user needs to provide the box sizes in pixels, otherwise raises an error. Because the angles cannot be converted to pixels without the knowledge of the frame resolution. This is why I also updated the prior notebook

Here is the output of the prior tutorial
Screenshot from 2025-04-18 02-04-00

and of the multiresolution tutorial
Screenshot from 2025-04-18 01-24-23

All clear for a review @pmelchior !

b-remy:
Instead of raising an error if `box_sizes` is not provided in pixels and the scene frame has no wcs, I am now printing a warning that the user should specify the `box_sizes` and use by default `box_sizes=[11, 17, 25, 35, 47, 61, 77]`.

b-remy:
@pmelchior, I have removed the source detection receipt from the multiresolution tutorial. It is now loading everything from a single fits file created and stored [here](https://github.com/astro-data-lab/scarlet-test-data/tree/main/scarlet_test_data/data/multiresolution_tutorial). Let me know if this works for you!

PR #132, opened 2025-04-16
adding option for explicitly set box sizes

SampsonML:
The prior is sensitive to box sizes. A nice way to explicitly set a box size, and have a best fit Gaussian initialization would help a lot.

### DIFF
--- a/docs/0-quickstart.ipynb
+++ b/docs/0-quickstart.ipynb
@@ -245,7 +245,7 @@
    "metadata": {},
    "source": [
     "`scene` contains the `frame` and the list `sources`. The first source is indeed a {py:class}`~scarlet2.PointSource`, which has a `morphology` model, namely the same {py:class}`~scarlet2.GaussianMorphology` as the PSF of the model frame.\n",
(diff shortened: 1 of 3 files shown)


pmelchior:
I would prefer a list of boxsizes instead. This would replace the minimum size, delta, and your fixed size. The idea is to run with the minimum size in the list and grow as needed from the other items in the list

SampsonML:
This could be done, but is there a way for the user to explicitly choose a box size from within the list. Say for whatever reason I want a boxsize of 32, and want to initialize from the Gaussian moments routine. This is mainly what I was after

pmelchior:
Sure. In your case, I'd imaging you provide `boxsizes=[32, 64]`.

SampsonML:
I see, I have implemented this here, removing min_size and delta_size.

pmelchior:
This PR changes the default box sizes to prefer the small end. It's better for isolating sources in absence of a monotonicity constraint. For NN priors it's possibly better to choose larger boxes.

PR #162, opened 2025-06-12
Including ruff linting and styling

drewoldag:
All source and test files have now been edited to pass ruff linting and style checks. 

Most of the changes are in one of the following categories:

1. Variable name updates
2. Docstring line length too long or missing
3. Use ternary operators in place of if/else
4. Imports either too loose, or not at the top of the file

Specifically for the docstrings, in the places where I felt like I really couldn't provide a good placeholder, I've created github issues for others to step in an provide the information. 

Overall, while a lot of files have been touched, ideally, there is no change in the functionality of the code. All the tests should still pass, all the notebooks should still run, the code will just be cleaning and more easily maintained.


src/scarlet2/bbox.py
```diff
@@ -111,43 +112,38 @@ def insert_into(self, image, sub):
         return image
 
     def get_extent(self):
+        """Return the start and stop coordinates."""
```
drewoldag, line 115:
```suggestion
        """Return the start and end coordinates."""
```

Issue #88, opened 2024-11-11
diagonalized fast Fisher matrix


SampsonML:
Move the implementation of the Hessian vector product approximation of the Fisher matrix into a standalone, runable location (maybe `utils.py`)
https://github.com/pmelchior/scarlet2/blob/9360b42c50ee20e364b66916840e50685bde4de7/scarlet2/plot.py#L555

This will include the `log_like` implementation (https://github.com/pmelchior/scarlet2/blob/9360b42c50ee20e364b66916840e50685bde4de7/scarlet2/plot.py#L534) 

as well as the neural gradient (https://github.com/pmelchior/scarlet2/blob/9360b42c50ee20e364b66916840e50685bde4de7/scarlet2/plot.py#L526)

PR #164, opened 2025-06-13
removed hallucination plot; closes #141

### LINKED ISSUE #141: plot.sources(show_hallucination=True) fails for PointSource object
In the `neural.ipynb` notebook, I attempted to run:
```
plot.sources(scene_,
             norm=norm,
             observation=obs,
             show_model=True,
             show_rendered=True,
             show_observed=True,
             show_spectrum=False,
             add_markers=False,
             add_boxes=True,
             show_hallucination=True,
             )
plt.show()
```

The scene included a PointSource, and the following error was produced:
```
AttributeError: 'PointSource' object has no attribute 'get_parameters'
```

pmelchior:
This PR removes the hallucination plot from the `plot.sources` method. The previous code is outdated because our treatment of `Parameters` has changed. But moreover, this functionality should be represented at the level of individual parameters, not sources (a source can have multiple parameters with different priors).

Re-integrating is linked to #88 because we need the function to perform the Fisher matrix computations by means of Hessian-vector product anyway: #88 needs it only for the log-likelihood function, and the hallucination score needs the HVP of the log prior in addition.


scarlet2/plot.py
```diff
@@ -657,7 +656,6 @@ def sources(
         norm=None,
         channel_map=None,
         show_model=True,
-        show_hallucination=False,
```
drewoldag, line 660: Will want to remove `show_hallucination` from the docstring as well.

drewoldag (approved):
Looks reasonable to me. Only one minor comment about updating the doc string.

scarlet2/plot.py
pmelchior: I thought I had done that...

PR #175, opened 2025-06-27
Scaffolding code for validation

drewoldag:
## WIP - Scaffolding for validation

Implementing some early scaffolding code for validation with the assumption that we want validation in three places:

### 1) During input
I believe this means just after instantiating a `Observation` object. The scaffolding would support two modes:
```
obs = Observation(data, weights, ..., check_observation=True)  # check_observation is False by default
``` 
or
```
from scarlet2.validation import check_observation

obs = Observation(data, weights, ...)
check_observation(obs)
```

### 2) After initialization of sources
The scaffolding would again support two modes: 
```
Source(center, spectrum, morph, check_source=True)
```
or
```
from scarlet2.validation import check_source

source = Source(center, spectrum, morph)
check_scene(source)
```

### 3) Output failures
I believe this means examining what has happened after `scene.fit()`. Again two modes would be supported:
```
scene.fit(obs, parameters, ..., check_fit=True)  # check_fit is False by default
```
or
```
from scarlet2.validation import check_fit

scene_ = scene.fit(obs, parameters, ...)
check_fit(scene_)
```

TODO:

- [x] Determine at which level to place the `check_source` function in source.py. `Source` is a child class of `Component` and there are a few other class in that module as well. I _think_ it makes sense to but the `check_source` function in the parent class, `Component`, and if so, it probably makes sense to rename it to `check_component`.

### DIFF
--- a/src/scarlet2/observation.py
+++ b/src/scarlet2/observation.py
@@ -14,7 +14,7 @@ from .renderer import (
     NoRenderer,
     Renderer,
 )
-from .validation import ValidationError, ValidationMethodCollector
+from .validation_utils import ValidationError, ValidationMethodCollector
 
 
 class Observation(Module):
@@ -46,7 +46,6 @@ class Observation(Module):
 
             validation_errors = check_observation(self)
             if validation_errors:
-                #! We can raise this as a ValueError or assign it to self.validation_errors
                 raise ValueError(
                     "Observation validation failed with the following errors:\n"
                     + "\n".join(str(error) for error in validation_errors)
--- a/src/scarlet2/observation.py
+++ b/src/scarlet2/observation.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import equinox as eqx
 import jax.numpy as jnp
 
@@ -12,6 +14,7 @@ from .renderer import (
     NoRenderer,
     Renderer,
 )
+from .validation import ValidationError, ValidationMethodCollector
 
 
 class Observation(Module):
@@ -144,3 +147,47 @@ class Observation(Module):
             ), "Renderer does not map model frame to observation frame"
         object.__setattr__(self, "renderer", renderer)
         return self
+
+
+class ObservationValidator(metaclass=ValidationMethodCollector):
+    """A class containing all of the validation checks for Observation objects.
+    Note that the metaclass is defined as `MethodCollector`, which collects all
+    validation methods in this class into a single class attribute list called
+    `validation_checks`. This allows for easy iteration over all checks."""
+
+    def __init__(self, observation: Observation):
+        self.observation = observation
+
+    def check_weights_non_negative(self) -> Optional[ValidationError]:
+        """Check that the weights in the observation are non-negative.
+
+        Returns
+        -------
+        ValidationError or None
+            Returns a ValidationError if the check fails, otherwise None.
+        """
+        if (self.observation.weights < 0).any():
+            return ValidationError(
+                "Weights in the observation must be non-negative.",
+                check=self.__class__.__name__,
+                #! Placeholder for a meaningful context
+                context={"observation.weights": self.observation.weights},
+            )
+        return None
+
+    def check_weights_finite(self) -> Optional[ValidationError]:
+        """Check that the weights in the observation are finite.
+
+        Returns
+        -------
+        ValidationError or None
+            Returns a ValidationError if the check fails, otherwise None.
+        """
+        if jnp.isinf(self.observation.weights).any():
+            return ValidationError(
+                "Weights in the observation must be finite.",
+                check=self.__class__.__name__,
+                #! Placeholder for a meaningful context
+                context={"observation.weights": self.observation.weights},
+            )
+        return None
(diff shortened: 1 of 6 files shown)


src/scarlet2/observation.py
```diff
@@ -25,8 +25,12 @@ class Observation(Module):
     """Metadata to describe what view of the sky `data` amounts to"""
     renderer: (Renderer, eqx.nn.Sequential) = eqx.field(static=True)
     """Renderer to translate from the model frame the observation frame"""
+    check_observation: bool = eqx.field(static=True, default=False)
+    """Whether to validation checks on the observation object. Default is False."""
```
pmelchior, line 29: I don't think this needs to be a class attribute. It's called at the end of __init__, so no need to remember it past that point.

src/scarlet2/scene.py
```diff
@@ -96,6 +102,10 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_value, traceback):
         Scenery.scene = None
+        if self.check_scene:
+            from .validation import check_scene
+
+            check_scene(self)
```
pmelchior, line 108: It's an elegant place for an automatic check. The only reason I don't quite like it is that if somebody adds one source after the initial creation of the scene, all existing sources would be checked again.

src/scarlet2/validation.py
```diff
@@ -0,0 +1,46 @@
+from typing import Optional
+
+from scarlet2.observation import Observation
+from scarlet2.scene import Scene
+
+
+def check_observation(observation: Optional[Observation] = None):
```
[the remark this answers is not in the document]
drewoldag: I did this just to keep things pretty flexible as this is starting to be built out. I think it is unlikely that we'll ever call these `check_*` functions without passing the class instance to be checked. I can clean this up and if we ever need it, it should be straightforward to put back in.

src/scarlet2/observation.py
drewoldag: Yep, you're right, I can remove this.

src/scarlet2/scene.py
```diff
@@ -25,13 +25,18 @@ class Scene(Module):
     """Portion of the sky represented by this model"""
     sources: list
     """List of :py:class:`~scarlet2.Source` comprised in this model"""
+    check_scene: bool = eqx.field(static=True, default=False)
+    """Whether to run validation checks on the scene after initialization. Default is `False`."""
```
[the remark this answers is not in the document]
drewoldag: This also makes sense, I can move things around here.

src/scarlet2/scene.py
drewoldag: Also very reasonable. It occurred to me that we could add a couple members to the `Observation`, `Scene`, or `Source` classes like: `self.validation_status: [UNCHECKED | PASSED | FAILED]` as well as a `self.validation_errors: list[ValidationDataclass]`.

But it seems that the class hierarchy with Equinox is such that updating class instance values isn't allowed. (Or rather, you can append to a list, but you can't update the status after creating the class instance.

For now I'm going to just return a list of validation error dataclasses from `check_*`, where each element in the list will have context of where it originated.

src/scarlet2/observation.py
drewoldag: Removed

src/scarlet2/scene.py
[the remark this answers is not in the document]
drewoldag: Removed this, and created a separate, standalone `check_scene` that just loops over the sources and calls the new `check_source` function.

src/scarlet2/validation.py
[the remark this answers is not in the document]
drewoldag: Removed the optional stuff.

drewoldag:
In the last commit I started to introduce return values. Right now what's happening is that we run through all the checks for the object, build up a list of `ValidationError` dataclass objects and then return that list. The current implementation of `ValidationError` is mostly placeholder until we have a better handle on what will be most useful for the user. 

Another approach that I looked into was setting a status flag and a list on the Observation, Scene and Source classes themselves. That way a given `Source`, for instance, could be validated and then hold onto the information about the validation. 

Two things came up, one, all those classes seem to be frozen dataclasses by way of Equinox, so updating a class attribute throws an error. And two, if the user did alter the class, then there's additional bookkeeping to make sure that the class is marked as "needs to be re-validated". 

The first of those two points seems to be the most painful. All that being said, it seems that the point of Equinox, and perhaps the expected usage of Scarlet2 is that the user would simply recreate a given invalid object, instead of trying to update it. @charlotteaward  @pmelchior - does that sound like the way you would expect users to interact with Scarlet2?

src/scarlet2/validation.py
```diff
@@ -1,46 +1,217 @@
...
+    @classmethod
+    def get_checks(cls):
+        """Get the list of all registered checks."""
+        if not hasattr(cls, "_checks"):
+            cls._checks = []
+        return cls._checks
+
+    def check(self) -> Optional[ValidationError]:
+        """Run the check on the object."""
+        raise NotImplementedError("Subclasses should implement this method.")
```
pmelchior, line 139: That feels somewhat heavy to me, and the sense that we need to create a class for each object we seek to test. I _thinkk_ we could do this simply with pytest-like functions.

src/scarlet2/scene.py
drewoldag: We've pulled the checking out of the `__exit__` method, and are focusing the checks on the sources at this point. The user has access to a `check_scene` function that will simply loop over the sources in the scene and check each one.

pmelchior (approved):

Issue #184, opened 2025-07-24
Parameter instances create jax warnings in the docs


pmelchior:
**Bug report**

The current docs show error outputs when constructing `Parameter`, e.g. [here](https://scarlet2.readthedocs.io/en/latest/0-quickstart.html#fit-the-model).
This seems to be a new problem (maybe from using a newer jax/jaxlib version). While it doesn't affect the fits, it's annoying and should be fixed.

```
Error processing fieldname 'node'. - __cuda_array_interface__ is only defined for NVidia GPU buffers.
Error processing fieldname 'node'. - descriptor '__weakref__' for 'Array' objects doesn't apply to a 'jaxlib._jax.ArrayImpl' object
Error processing fieldname 'node'. - arr.device_buffer has been deprecated. Use arr.addressable_data(0)
Error processing fieldname 'node'. - arr.device_buffers has been deprecated. Use [x.data for x in arr.addressable_shards]
Error processing fieldname 'node'. - JAX Arrays do not implement the arr.flat property: consider arr.flatten() instead.
Error processing fieldname 'node'. - x must be at least two-dimensional for matrix_transpose; got ndim=1
```

Issue #156, opened 2025-06-11
Get memory usage benchmarks for `scarlet2`


maxwest-uw:

maxwest-uw:
## Quickstart Guide
Input file size (compressed, h5 file) - 8kb
Scene + Model load in - ~200 mb
Max memory usage - 800 mb
4x memory increase during runtime

![Image](https://github.com/user-attachments/assets/ce9440d0-24cc-46c7-a65b-58432fc5ebee)

maxwest-uw:
## Timedomain
input data - around 250mb
max memory usage - 1.75gb

![Image](https://github.com/user-attachments/assets/0f452ba4-aba8-4064-9462-3b11dea17816)

maxwest-uw:
## Sampling
input data - around 300mb
sampling - around 600mb (doubles throughout this process)

![Image](https://github.com/user-attachments/assets/6f069069-e1b2-4b8f-bb01-b55aea18579c)

maxwest-uw:
## Priors
Input file size (compressed, h5 file) - 8kb
Scene + Model load in - ~200 mb
Max memory usage - 800 mb

![Image](https://github.com/user-attachments/assets/23e7c2da-51a1-4ff5-a151-2523b8655046)

Issue #120, opened 2025-04-08
check conditions for jit cache misses


pmelchior:
We have reasons to strongly suspect that processes that fit different scenes/observations trigger cache misses for the jitted `Scene._make_step` function. This issue is about diagnosing the problem.

JAX provides a utility function for cache misses: https://docs.jax.dev/en/latest/persistent_compilation_cache.html#examining-cache-misses
Let's find out if merely providing 
1. a different scene (with the same parameterization) but the same observation
2. a different observation with the same scene parameterization
triggers a cache miss.

I suspect it only happens with option 2. If so, we need to investigate how to avoid the cache miss for observations that are fixed-format and only differ by the PSF because that'll be the most common configuration for large-scale survey processing.

b-remy:
There seems to be a cache miss for the `_make_step` function in both cases, resulting in it being recompiled for each new fit. Here is the warning:

```
TRACING CACHE MISS at /home/ben/Documents/repos/scarlet2/scarlet2/scene.py:306 (fit) because:
  for _make_step defined at /home/ben/micromamba/envs/galsim/lib/python3.9/site-packages/equinox/_jit.py:37
explanation unavailable! please open an issue at https://github.com/google/jax
Compiling _make_step with global shapes and types [ShapedArray(float32[2]), ShapedArray(float32[1]), ShapedArray(float32[31,31]), ShapedArray(float32[1,128,128]), ShapedArray(float32[1,128,128]), ShapedArray(int32[]), ShapedArray(float32[1]), ShapedArray(float32[31,31]), ShapedArray(float32[1]), ShapedArray(float32[31,31]), ShapedArray(int32[])]. Argument mapping: [UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue, UnspecifiedValue].
Finished jaxpr to MLIR module conversion jit(_make_step) in 0.058747291564941406 sec
Finished XLA compilation of jit(_make_step) in 0.20924758911132812 sec
100%|████| 200/200 [00:00<00:00, 249.44it/s, loss=15199.57, max_change=0.004965]
```

don't know how to fix this yet, as it is not very explicit

pmelchior:
[...]

PR #188, opened 2025-07-29
Housekeeping

pmelchior:
This PR fixes three different problems.

1. It removed any occurence of equinox dataclass attributes that were defined with `eqx.field(init=False)`. That used to be the mechanism to define an attribute that is not present as argument in the default `__init__`. However, equinox new complains about this problem rather verbosely, and we always have custom inits for those ones anyway, so uses of `init=False` and subsequence `getattr` have been replaced by standard `self.attribute = attribute`.
2. I managed to remove the last occurrence of `eqx.field(static=True)` from the code base. It was for `Observation.renderer` and was used to paper over a different problem. Some of the renderers use slices to narrow the number of channels or the spatial box, but these cannot be hashed in python <3.12. JIT will not worked if static parameters cannot be hashed, so I cerated a hashable slice data class.
3. jax throws a lot of exceptions when a class attribute is requested that is deprecated. Since we seek to convert astropy units to pixel units in our parameters, we need to inspect every attribute. That causes a lot of spam (#184). I silence these.

### DIFF
--- a/docs/howto/multiresolution.ipynb
+++ b/docs/howto/multiresolution.ipynb
@@ -22,11 +22,10 @@
     "\n",
     "# Import Packages\n",
     "import jax.numpy as jnp\n",
+    "import scarlet2\n",
     "from astropy.coordinates import SkyCoord\n",
     "from astropy.table import Table\n",
-    "from astropy.wcs import WCS\n",
-    "\n",
-    "import scarlet2"
+    "from astropy.wcs import WCS"
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


pmelchior:
There's a possible side effect of items 1 and 2 above. Since argument that are not initialized or declared static can interfere with `grad` and `jit`, this might be the cause for our cache misses (#120) or memory consumption (#112, #156).

@maxwest-uw Can you review this PR and try your memory test on this branch?

maxwest-uw (approved):

PR #189, opened 2025-07-30
Add questionnaire models

smcguire-cmu:
Adds the pydantic models for use in the input questionnaire

### DIFF
--- a/src/scarlet2/questionnaire/models.py
+++ b/src/scarlet2/questionnaire/models.py
@@ -2,23 +2,35 @@ from pydantic import BaseModel, Field
 
 
 class Template(BaseModel):
+    """Represents a template for code replacement in the questionnaire."""
+
     replacement: str
     code: str
 
+
 class Answer(BaseModel):
+    """Represents an answer to a question in the questionnaire."""
+
     answer: str
     tooltip: str = ""
     templates: list[Template]
     followups: list["Question"] = Field(default_factory=list)
     commentary: str = ""
 
+
 class Question(BaseModel):
+    """Represents a question in the questionnaire."""
+
     question: str
     answers: list[Answer]
+
 
 Question.model_rebuild()
 Answer.model_rebuild()
 
+
 class Questionnaire(BaseModel):
+    """Represents a questionnaire with an initial template and a list of questions."""
+
     initial_template: str
     questions: list[Question]
--- a/src/scarlet2/questionnaire/models.py
+++ b/src/scarlet2/questionnaire/models.py
@@ -14,7 +14,7 @@ class Answer(BaseModel):
     answer: str
     tooltip: str = ""
     templates: list[Template]
-    followups: list["Question"] = Field(default_factory=list)
+    followups: list["Question"] = Field(default_factory=list)  # Forward reference to Question
     commentary: str = ""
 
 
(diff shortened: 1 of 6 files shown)


src/scarlet2/questionnaire/models.py
```diff
@@ -0,0 +1,36 @@
...
+    code: str
+
+
+class Answer(BaseModel):
+    """Represents an answer to a question in the questionnaire."""
+
+    answer: str
+    tooltip: str = ""
+    templates: list[Template]
+    followups: list["Question"] = Field(default_factory=list)
```
drewoldag, line 17: Just my own naivety, why is `"Question"` in quotes and `Template` isn't?

src/scarlet2/questionnaire/questions.yaml
```diff
@@ -0,0 +1,91 @@
...
+    followups:
+    - question: Do you wish to model transient or variable sources across multiple epochs?
+      answers:
+      - answer: "Yes"
+        templates:
+          - replacement: code
+            code: "channels=[(‘g’,0)]
+                   \nchannels=[(‘g’,1’)]\
+                   \n\n{{obs}}"
+        commentary: "For each image, the channels should be a list of tuples where each tuple provides the filter and an epoch identifier which is unique for each image of a particular filter. For example, you might have two g-band images from different times, which would be initialized separately with channels=[(‘g’,0)] and channels=[(‘g’,1’)]. You might have two other r-band images taken at different times again. These would be initialized with channels=[(‘r’,0)] and channels=[(‘r’,1’)]. You may wish to take note of any epochs where the transient is known to be ‘off’ - you can add this constraint later if desired. Each image can be initialized"
```
drewoldag, line 24: Has this commentary been cut off at the end? Also, what is the single quote for in `channels=[('g',1']` and `channels=[('r',1')]`?

src/scarlet2/questionnaire/questions.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - answer: "No"
+        templates:
+        - replacement: code
+          code: "channels=['g', 'r', 'i', 'z', 'y']\n\n{{obs}}"
+        commentary: "For each image, the channels in the image can be specified as a list of the filters"
```
drewoldag, line 29: Same as line 11?

```suggestion
        commentary: "For each image, the channels in the image can be specified as a list of the filters, like in the above example."
```

src/scarlet2/questionnaire/questions.yaml
drewoldag, line 1: Not sure how finalized this set of questions is, so I don't want to go overboard with tiny nits.

drewoldag (approved):
The substance of the PR looks good to me. I left a couple questions on the .yaml file. Including one indicating that I'm not sure how finalized it really is, so I didn't want to get too fine grained calling out out punctuation and such.

src/scarlet2/questionnaire/models.py
smcguire-cmu: Ah, that's because the `Question` class hasn't been defined yet. It's also why `model_rebuild` needs to be called to update the models after `Question` is defined. I've added a comment to explain that.

src/scarlet2/questionnaire/questions.yaml
smcguire-cmu: Yeah, I think it will keep changing quite a bit as it continues to be worked on.

PR #206, opened 2025-08-21
WIP - Writing up documentation for validation.

drewoldag:
Primarily focused on dev guide portion for this commit - aside from small tweaks this portion should be finished - happy to take early comments. 

I've only dropped random thoughts into the user guide portion, so I'll fill out with the next commit.

### DIFF
--- a/docs/2-validation.md
+++ b/docs/2-validation.md
@@ -1,0 +1,277 @@
+# Validation
+
+## Automatic validation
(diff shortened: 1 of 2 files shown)


pmelchior (approved):
Looks good to me. We should use the same structure also for documenting the questionnaire.

maxwest-uw (approved):

PR #208, opened 2025-08-22
Clean up docstrings and type hints in validation files

drewoldag:
There are no changes to logic in this PR, it is strictly correcting type hints and docstrings to reflect what the code is actually doing. Also updated a couple variable names in validation.py for the same purpose.


charlotteaward (approved):
Looks good!

Issue #209, opened 2025-08-26
Slicing errors when modeling multi-observation scene


charlotteaward:
I'm currently modeling 3 Euclid images with size (1,1000,1000) and PSF size (1,33,33). Initialization all works fine and the images look good, but I get this slicing error during scene.fit:

```
Traceback (most recent call last):
  File "/u/cw152/DDP3/queryEuclidcutouts.py", line 230, in <module>
    scene_ = scene.fit(obs, parameters, max_iter=stepnum, e_rel=1e-7, progress_bar=False)
  File "/u/cw152/.conda/envs/my.anaconda/lib/python3.13/site-packages/equinox/_module/_prebuilt.py", line 33, in __call__
    return self.__func__(self.__self__, *args, **kwargs)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/cw152/.conda/envs/my.anaconda/lib/python3.13/site-packages/scarlet2/scene.py", line 314, in fit
    scene, loss, opt_state, convergence = _make_step(
                                          ~~~~~~~~~~^
        scene, observations, parameters, optim, opt_state, filter_spec=filter_spec
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/u/cw152/.conda/envs/my.anaconda/lib/python3.13/site-packages/equinox/_jit.py", line 209, in __call__
    return _call(self, False, args, kwargs)
  File "/u/cw152/.conda/envs/my.anaconda/lib/python3.13/site-packages/equinox/_jit.py", line 263, in _call
    marker, _, _ = out = jit_wrapper._cached(
                         ~~~~~~~~~~~~~~~~~~~^
        dynamic_donate, dynamic_nodonate, static
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/u/cw152/.conda/envs/my.anaconda/lib/python3.13/site-packages/scarlet2/scene.py", line 490, in _make_step
    loss, grads = filtered_loss_fn(diff_model, static_model)
                  ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #215, opened 2025-08-28
Change `min` to `max`  FFT shape in resampling renderer

b-remy:
This PR is addressing #209 

Using the `max` instead of `min` makes the convolved Fourier image as big as the biggest of the model or PSF Fourier image, and the following `_trim` error always working.


charlotteaward:
Confirming that this fixes the problem and it runs correctly on larger images.

PR #219, opened 2025-09-05
Add Questionnaire Documentation

smcguire-cmu:
Adds the questionnaire documentation page. I put this as the second thing in the tree before the validation since I thought that's where it fit best, but happy to change that around.

This also adds the feedback link at the end of the questionnaire and in the docs, and finishes up adding the last few questions in the document to the `questions.yaml` file. I also separated out the imports to their own section at the top of the templates as the questions get answered.


charlotteaward (approved):
The docs are really clear :) All looks great to me!!

PR #221, opened 2025-09-08
Add Back Navigation and Saving/Loading to questionnaire

smcguire-cmu:
Adds the feature to click on previous answers to navigate back to them, and the ability to save and load questionnaire answers.


drewoldag (approved):
Overall this looks really good to me. I pulled the branch and played around with it. It's works as advertised. 

The only additional comment that I have is the input variable `save_path`. It tripped me up. Though to be fair, I read the documentation a while before I actually pulled the code and tried to use it. Based on the variable name I was expecting to provide the full path with file name, which produced an error. Then I tried just the file name (still without looking back at the docs) and it produced an error. Then I read the docs, and got it right. 

Perhaps a name like `save_directory` would have pushed me in the right direction from the beginning.

charlotteaward (approved):
This looks perfect!!

PR #225, opened 2025-10-05
Correct rotations even when improper

pmelchior:
WCS transformation matrices are often used to create a left-handed coordinate system to get the conventional North up, East left orientation. That amounts to an improper rotation from the Euclidean pixel grid. 

To state the convention we use in scarlet2 explicitly: A counter-clockwise rotation by an angle $\theta$ is given by the matrix

$$
R = \begin{pmatrix}
\cos \theta & -\sin\theta\\
\sin\theta & \cos\theta
\end{pmatrix}
$$

An _improper_ rotation is a combination of a rotation and a flip:

$$
\overline{R} = \begin{pmatrix}
\cos\theta & \sin\theta\\
\sin\theta & -\cos\theta
\end{pmatrix} =
\begin{pmatrix}
\cos \theta & -\sin\theta\\
\sin\theta & \cos\theta
\end{pmatrix}
\begin{pmatrix}
1 & 0\\
0 & -1
\end{pmatrix}
$$

(this uses the convention of a flip in y, but we can adopt it over a flip in x without loss of generality). 

So, if we have an improper rotation, the convention for the angle becomes ambiguous. To make it unambiguous, to get the angle from a transformation matrix $M$, one has to ensure it is a proper rotation. We therefore now test the determinant and do something like this to get the angle:
```python
det = jnp.linalg.det(M)
if det < 0:
    F = jnp.diag((1, -1))
    R = F @ M   # formally F^-1 @ M, but F = F^-1
else:
    R = M
angle = jnp.arctan2(R[1, 0], R[0, 0])
```

Earlier version of our code did not check or account for that because
* our example of multi-resolution fitting in the docs is rectified, so the angle was 0 and the orientations of both images are the same (left-handed), so no need to do anything.
* our test case in `test_resampling` was wrong. It attempted to test a 90-degree counter-clockwise rotation but failed twice to succeed: it ignored the RA-flip in the HST WCS and incorrectly applied `jnp.rot90`, which by default rotates in axis order  `(0,1)`, which is opposed to the conventional rotation from the positive x-axis (that's axis=1, not 0). The test image was therefore rotated 90 degrees _clockwise_. With the first mistake, the test succeeded, even though the internal treatment was incorrect.

This PR fixes this problem, makes minor changes to `Moments` to ensure consistency, and removes `set_spectra_to_match` because it is brittle in multi-resolution cases, slow, and not particularly useful.

### DIFF
--- a/src/scarlet2/frame.py
+++ b/src/scarlet2/frame.py
@@ -55,10 +55,7 @@ class Frame(eqx.Module):
         float, astropy.units.quantity.Quantity
             Pixel size in units of the WCS sky coordinates
         """
-        if self.wcs is not None:
-            return get_scale(self.wcs)
-        else:
-            return 1
+        return get_scale(self.wcs)
 
     def get_pixel(self, pos):
         """Get the sky coordinates from a world coordinate
@@ -73,7 +70,6 @@ class Frame(eqx.Module):
         pixel coordinates in the model frame
         """
         if isinstance(pos, SkyCoord):
-            assert self.wcs is not None, "SkyCoord can only be converted with valid WCS"
             wcs_ = self.wcs.celestial  # only use celestial portion
             pixel = jnp.asarray(pos.to_pixel(wcs_), dtype="float32").T
             return pixel[..., ::-1]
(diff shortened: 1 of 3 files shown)


pmelchior:
Btw, some useful reading on (im)proper rotation matrices [here](https://textbooks.math.gatech.edu/ila/complex-eigenvalues.html) and [here](https://scipp.ucsc.edu/~haber/ph116A/Rotation2.pdf)

PR #239, opened 2025-12-10
Correlated observation

pmelchior:
This PR brings two major changes:
* Correlated noise (e.g. in a coadd) can treated by measuring its correlation function, converting to the power spectrum and computing the chi-squared in Fourier space (#195 )
* The multi-resolution fit can now by sped up by resampling the data first to the same pixel grid as the model frame. This correlates the pixel values, but we can treat that by the same mechanism.

So, the standard recommendation for multi-observation fitting in the docs is now :

```python
obs_resampled = sc2.CorrelatedObservation.from_observation(obs, resample_to_frame=model_frame)
scene.fit([obs_highres, obs_resampled], parameters, ...)
```

Plotting, init, fitting, and sampling functions have been slightly modified to work with resampled observations. I expect that only high-precision science cases (such as cosmic shear) will need more accurate treatment than the internal Lanzcos resample (but calibration schemes like metadetection should be just fine even with resampling).


charlotteaward:
@pmelchior It's running smoothly. At the moment for my test case and the parameter step sizes that worked alright previously, it gets here after 3000 steps:

VIS_0_fit_Rubinonly
g_1_fit_Rubinonly

And a minor note, at the moment when plotting a single observation that has the channels listed as tuples, we still need to set split_channels=True otherwise we get:

```
Traceback (most recent call last):
  File "/u/cw152/DDP3/modelEuclidRubin_correlatedTD2.py", line 450, in <module>
    ani = scarlet2.plot.scene(scenes, observations[0], show_model=True, show_rendered=True, show_observed=True, show_residual=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/u/cw152/scarlet2/src/scarlet2/plot.py", line 961, in scene
    name = "".join(observation.frame.channels)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: sequence item 0: expected str instance, tuple found

pmelchior:
Ah, yes. I knew of this problem. Fix is coming.

As for the example, I don't like the diagonal hatching. This looks like the 2D power spectrum is quite funky. Can you run it without using the correlated observation for VIS, please.

PR #245, opened 2026-02-24
Param as dict

pmelchior:
This PR brings significant changes under the hood in who parameters are stored and treated. In particular:

* `Parameters` now derive from `dict`, so allow lookup like `parameters["spectrum_0"]`.
* The global namespace hold a `parameter_registry`, which uses the memory location of a module instance to hold the parameters of that instance. This entry is created by the init function `Parameters(some_module_instance)`. The parameter set can be retrieved by calling the property `Module.parameters`. `Module` also allows for `set` and `get` of the parameter values.
* The new free functions `infer.fit` and `infer.sample` now accept parameters from multiple modules and internally combine them into a dynamically created eqx.Module that contains attributes for every key/value pair in the parameters dict. Only this new module of values get optimized/sampled, which massively reduces the overhead per iteration as it avoids copying non-optimized data structures.

The main purpose is to allow Parameters to be defined not just for `Scene`, but for any other `Module`, in particular `Observation`. As an example, we can adjust the astrometric offset by parameterizing the `.shift` attribute in the `ConvolutionRenderer`:

```python
with sc2.Parameters(obs_hsc):
    sc2.Parameter(obs_hsc.renderer[1].shift, name="shift", stepsize=1e-3 * u.arcsec)
```

A detailed example is given in the multi-observation howto.

As a consequence of these changes, the signature of the `fit/sample` functions does not require the `parameters` argument (it's held already by the module that gets optimized). This shorter call signature will do and is now used in the docs: 

```python
with sc2.Parameters(scene): # note: no "as parameters"
    ...
scene.fit(obs, max_iter=1000) # note no parameters argument
```

While the changes are quite substantial, scripts that are based on the v0.3 API should all still work.

### DIFF
--- a/docs/howto/priors.ipynb
+++ b/docs/howto/priors.ipynb
@@ -138,11 +138,7 @@
    "source": [
     "## Define Parameters with Prior\n",
     "\n",
-    "We use the same fitting routine as in the Quickstart guide, but replace `contraints.unit_interval` with `prior=prior` in the Parameter containing the source morphologies. As we will be running an optimizer, not a sampler, we still need to define `stepsize` for this parameter. It is useful to maintain small step sizes for the morphology updates because large jumps can lead to unstable prior gradients.\n",
-    "\n",
-    "```{info}\n",
-    "You cannot set a `prior` and a `constraint` for the same parameter. The reason is that a constraint will be treated by transforming the variable in question, but then it's unclear if `prior` describes the variable before or after the transformation.\n",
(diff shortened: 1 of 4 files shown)


charlotteaward (approved):
Confirm that it runs on test data and seems to fit the shifts well. Don't have any finalized fits yet but here's what it all looks like after 1000 iterations:
Rubin:
Image
Roman:
Image

PR #293, opened 2026-09-10
Cleanup

pmelchior:
## Change Description

Provides validation tests with output modes "off", "on", "verbose". The default "on" mode, reports the result of the test, but not the context and is easier to parse. "verbose" includes the context.

This PR also removes the questionnaire. Doing so reduces code dependencies and avoids stale questionnaire recommendations. In the age of LLMs with good code docs, it doesn't provide any benefit.


pmelchior:
@charlotteaward I pinged you for this PR just FYI. I think both changes in this PR are improvements, but let me know if you disagree.

